PR #18, opened 2015-01-26
Cancel damage if unlogged-in player is attacking other entities

orkun1675:
Right now players can hit other players before logging in, causing serious issues with plugins like CombatTag, PvPManager etc.
We should cancel this in a simple EntityDamageByEntity listener.

Did not test the code.


Xephi:
thanks :)

orkun1675:
Can you compile a dev version and upload it to Atlassian so I can test please?

PR #65, opened 2015-06-17
Fixed Problem with Dependencies conflict

sgdc3:


sgdc3:
Ups I think this may not work! Waiting for travis result

Xephi:
Why don't use the official Essentials repo ?
http://repo.ess3.net/content/groups/public

sgdc3:
Because Essentials is outdated, EssentialsX have a lots of fix and enhancements.

Xephi:
We only need the api to compile with, is it really important? >.<

Xephi:
Build Success when change from EssentialsX to Essentials ^^"

sgdc3:
I didn't check if they have fixed issues about the API

sgdc3:
Ok, i will check if there are differences in the API

sgdc3:
You're right, if there aren't issues it's useless.

PR #89, opened 2015-07-05
Enhanced Password Regex

Maxetto:
We are now allowing every ASCII character except for space (0x20 - 32). Those are the characters allowed in passwords.

Note: You may want to check if in config a single slash is OK for excluding a character, or if we need 2 like in the code.


Xephi:
let people us their own, the base one is simple and suffisent

Maxetto:
Not really much efficient, for exemple I couldn't register with that REG_EX, and since many people is not experienced with it, we should allow everything that is possible.
So i checked on google what are the characters allowed in passwords and I found that.

PR #112, opened 2015-07-10
Language updates, Admin Command rework and typo fixes

Maxetto:
I've added Turkish and Basque translations (Sent on Bukkit Dev),
I've changed Italian translation once again,
I've updated ReadMe so it shows every language that is currently on the plugin (should be updated on Bukkit Dev and Spigot too),
I've reworked Admin commands (We now check for unsafe passwords even there at changepassword and register, we now check if the email is valid, we now send a message everytime, I've merged "removeposition" with "purgelastpos" as they seem to be the same, I've added missing commands to the help (and moved them in category-like commands) and I've changed some parts of code to match the structure of the latest ones).

Yet, there are 2 things that I don't know how to do:
//TODO: Still needs to block output if doesn't have the permission for "Usage:" messages.
//TODO: Add a way to replicate the "/authme" command when "/authme args[0]" is not in the list.)

I moved register password checks with the others 2 (at the top, in the pre-register) and eased them out,
And finally, I've fixed every IP Typo (was: 198.18.0.1, now is: 192.168.0.1)

P.S. As I forgot to merge before starting on this, it's impossible to automatically merge, but I've pushed your commits later, so you can just safely copy and paste the code.


sgdc3:
@Maxetto Merge my PR on your Fork! I fixed all the conflicts!

src/main/java/fr/xephi/authme/commands/AdminCommand.java
```diff
@@ -97,7 +103,11 @@ public boolean onCommand(CommandSender sender, Command cmnd, String label,
 
         if (args[0].equalsIgnoreCase("purge")) {
             if (args.length != 2) {
-                sender.sendMessage("Usage: /authme purge <DAYS>");
+                sender.sendMessage("Usage: /authme purge <days>");
+                return true;
+            }
+            if (Integer.parseInt(args[1]) < 30) {
```
Xephi: Nope! I don't agree with that change :) let users choose

Xephi:
I don't agree with one change and don't forget to be care with changing ip to 192.168.0.1, cause database may contain 192.68.0.1

Maxetto:
Well, this is not used anymore. It was precedently used for sessions, but now they work differently, so we can change them safely.

src/main/java/fr/xephi/authme/commands/AdminCommand.java
Maxetto: I do agree, I got once my database full purged by an admin that did /authme purge 1, as he tought he was working inversly (would only clear data till 1 day before).
This is a Safe measue for mistype.
You may want to set this to one week or one day if issued by console ;)

sgdc3: @Xephi @Maxetto we can add an option: minPurgeDays with default 30 ;)

Xephi: Yep @sgdc3 , nice idea

Maxetto: I disagree with that. I mean, if you need to clear database before 30 days you can just delete your database straight. Why would you purge everything that was created before 30 days?
It's kind of a safe hard limit, where you can't go under.
I still would suggest to keep 30 days for the client and 1 week-1 day for the console.

sgdc3: In this case I agree with @Maxetto

PR #117, opened 2015-07-13
Fixed duplicated mail api in the shaded jar

sgdc3:
@Xephi Please, check if this is correct ;)


Xephi:
No problem on sending mail appear?

PR #127, opened 2015-07-16
Readme and text stuff

sgdc3:


src/main/java/fr/xephi/authme/AuthMe.java
```diff
@@ -1,782 +1,781 @@
-package fr.xephi.authme;
```
sgdc3: Sorry, the git client on my mobile is crazy -_- I changed only some wrong console strings...

Issue #223, opened 2015-09-28
Registration exploit


JunkyBulgaria:
Whenever MySQL(phpBB or anything else probably) is used as user storage, they can re-register instead of logging in.

Example: I have already registered, when i type "/register 123456789" it will say that i've successfully registered and can play.

games647:
reproducible for flat file storage

sgdc3:
damn, we are going to fix this asap! D:

games647:
I guess I found it

sgdc3:
Fixed version here: http://ci.xephi.fr/job/DEVBUILD-ONLY/lastSuccessfulBuild/artifact/target/AuthMe-5.1-SNAPSHOT.jar @JunkyBulgaria

JunkyBulgaria:
yep, works fine

sgdc3:
Can you close the issue?

PR #224, opened 2015-09-28
Fix duplicate registers

games647:
Related issue #223 

If the list of unsafe password in the configuration isn't empty it doesn't check if the account already exists. It just skips the other checks and continue the register process like everything else is checked.

This pr also fixes a invalid repository link where it's impossible to download the dependency from Essentials.


sgdc3:
Can you close this? @games647

PR #225, opened 2015-09-30
Replacing old inventory protecting with safe packet modifications usi…

games647:
…ng ProtocolLib.

Instead of clearing the inventory of players and storing it's contents in a file, we now prevent
the server from sending the inventory packet if the player is not logged in. The player will
see a empty inventory, but has still his items stored on the server. Therefore we don't 
need to modify the player's inventory and we won't make any inventory corrupted.

Fixes Xephi/AuthMeReloaded#203, 
Fixes Xephi/AuthMeReloaded#193, 
Fixes Xephi/AuthMeReloaded#191, 
Fixes Xephi/AuthMeReloaded#148

I replaced the old ProtectInventoryEvent and similar events with a representation of the player's inventory at the current stage. Modifications to that event will change the actual player inventory. This keeps the events backwards compatible.

**This pr needs further testing before merging.**


games647:
What should happen to the events? I mean, should the inventory be cleared if the StoreInventoryEvent or the RestoreInventoryEvent is cancelled?

sgdc3:
i think it should keep the inventory, authme doesn't need to edit the inventory anymore, we have only to keep the api set inventory method.

games647:
Maybe plugins actually uses with method. Wouldn't we break then the functionality of them?

sgdc3:
We provide compatibility only with the api, developers shouldn't use code outside of it

games647:
Events are part of the api

sgdc3:
Just keep them but deprecated, call them when the inventory is resent

games647:
Ok, then it's ready to be merged

sgdc3:
Can you close the PR? Already merged in the team repo :)

PR #398, opened 2016-01-01
Fix player not being able to move after registration

theJ3H:


DNx5:
Then, what's the point of this check? https://github.com/Xephi/AuthMeReloaded/commit/efe96b16181f8687f58fa0307f688b377826f096#diff-f457381723d83e20bc3dddf214d34d8bR120
that's not a fix.

theJ3H:
Well, the player is successfully logged in, so they should be able to move right?
If the walkspeed is set to 0.0 after registration the player won't be able to move.
On my server I want to use the removeSpeed feature but I don't want players to have to reconnect after registration because they can't move.

DNx5:
I think the speed restore should be done in login process.

sgdc3:
Yeah, not a fix

DNx5:
also, if you want to open a pull request. do it at team repo
https://github.com/AuthMe-Team/AuthMeReloaded

PR #526, opened 2016-02-14
Fix Django crypt

bakatrouble:
There were fixed 15k iterations originally, but with Django 1.9.2 (idk about previous versions) all passwords are encrypted with 24k iterations by default, so AuthMe was unable to verify even correct passwords. Now iterations are being read from hash string.


ljacqu:
This is unfortunately a PR for Xephi:master, but we develop AuthMe at AuthMe-Team:master, which is almost 500 commits ahead ^^ I've cherry-picked your commit, did some adjustments and created a PR for AuthMe-Team:master—https://github.com/AuthMe-Team/AuthMeReloaded/issues/102

Please review if you agree with it

sgdc3:
Merged on the other side ;)

Issue #938, opened 2016-09-04
AntiBot: find better implementation for users in time span


ljacqu:
The AntiBot class checks that in the last 5 seconds not more than x players have joined. This is being solved with threads, so a task with a wait time is scheduled to decrease the counter again. Find a better way of doing this without scheduling threads.

Platinteufel:
With the current implementation AntiBot can also enable after a server restart, when all players try to rejoin again. Maybe there should be an option that AntiBot only triggers on unregistered players? I think you don't have to check regular players.
If I get a botattack the probabilty that it's not a unknown player is very low.

ljacqu:
@Platinteufel this is the intended behavior of AntiBot (only fire if X *unregistered* players are observed). Please open another issue if this is not the case

Issue #937, opened 2016-09-04
Allow AuthMe to run in sync or async (configurable)


ljacqu:
Via Gitter with @sgdc3—
Add a configuration that will make AuthMe run synchronously instead of asynchronously.

- See changes in the sync branch for what is affected
- Async scheduling usually runs through the BukkitService -> I would expect we only need a few modifications in there (checking an additional setting)
- Note that _some_ things will still have to be in async (e.g. converters), so we need to create another method in BukkitService for always being in async

ljacqu:
@Sitmobedaf, one open question about this, did you expect code to run in another _sync thread_ or just run the code immediately? Right now we do the latter and it's unclear to me what is better.

Sitmobedaf:
I use 3.1.1 with fixed getOnlinePlayers and single session. I do not care what you do now, sorry.

sgdc3:
@Sitmobedaf Fine, bye.

Sitmobedaf:
Yes, it works synchronously and the load on the server attacks is almost there, it loads the server is less than SkinsRestorer, PermissionsEx, he BukkitServer. Authorization plugin does not have to be a monster - this is the direction in which you move, and you piatetes subdue him (asynchronous mode). In addition, you can not operate some of the functions of such a purge. But maybe I will come back to you. With you it was nice to talk to, good luck!

Issue #930, opened 2016-09-02
Add captcha to register? Solution for spambot


Typhome:
We're having problem with spambots that can register and then full spam server chat.
I thought that register with captcha would help to prevent this... IP ban is not helpful here, because it uses proxy IPs (+ dynamic IPs etc).

Picture of spambot doing task: https://i.gyazo.com/ab3a705b350db0144bba369351e03d68.png

sgdc3:
Noncheatplus? xD

sgdc3:
Captcha on register would annoy players, we can add it but i don't thibk it would be good dor your server.

ljacqu:
Agree, but on the other hand registering will only be done once per player :)

sgdc3:
@ljacqu I would close this as wontfix, is that ok for you?

ljacqu:
It's a reasonable feature request. Why are you against it?

sgdc3:
> but on the other hand registering will only be done once per player :)

You said that, sorry, i'm not against that, but we have to move it to the next milestone

ljacqu:
Sorry, it's been a stressful week for me xP

sgdc3:
> Wow, you're quoting me totally out of context. It was a reply to you saying we shouldn't do it because it will annoy people

Sorry i think i didn't understand the context at all, sorry ;)

ljacqu:
The sorry's all mine; I didn't have to get so worked up ^^ Sorry :)

ljacqu:
I'm stuck at the moment because when a user connects to a server I need to use the RegisterCaptchaManager in BungeeService to check if a captcha is required (to set the appropriate message), but at the same time I need to use BungeeService inside RegisterCaptchaManager to update the captcha message when the code check has failed (because the code then changes).

So either I split services, introduce a mechanism to have circular dependencies, or I go the cheap way and just set a message on login that doesn't contain the captcha code. I'm not happy with any of those approaches.

ljacqu:
Hi @sgdc3 @games647 @HexelDev, although I'd prefer not to have any circular dependencies I think this might be the best approach for this case here. I don't see any natural way to split the services. What do you think?

games647:
@ljacqu Does the captcha service really needs to have BungeeCord access? Shouldn't there only be two cases:

1) player is logged in by the previous server -> force login without captcha
2) player is not logged in -> require register with an independent generated captcha

or do I miss something?

ljacqu:
[...]

Issue #830, opened 2016-07-04
[Dev] (A)sync processes should be initiatied with more specific methods


ljacqu:
The management class should offer a "force register" method for registering a player instead of only one method with some "magic booleans" or other arguments that are empty in some use cases. This not only applies to registration process.

RegisterAdminCommand duplicates a lot of logic, which is bad. It should get an entry point into the Register async task like the other paths. Otherwise this keeps leading to bugs when something is changed in the register process and the /authme register command is forgotten.

This story should also remove the `boolean autoLogin` handling with a more elegant feature (e.g. multiple methods to fire the registration event).

PR #940, opened 2016-09-04
5.2 [WIP]

sgdc3:


sgdc3:
@ljacqu could you paste your todo list from gitter?

ljacqu:
In regards to 5.2, I'd say it would be nice to include #472, so I'll do that in the course of the week. 
#830 is something I've started, and so should continue.
Then, #937 or #938 would definitely be cool.
Finally, #930 might require some tinkering with the injector if we want to reuse the captcha manager, so that's something I probably have to look at

PR #1164, opened 2017-04-13
Update Dutch messages

timvisee:
I've updated all Dutch messages, and added the TODOs left in the messages file.


sgdc3:
Thank you ;)

Issue #1150, opened 2017-03-27
Add hashing support for Argon2


mat41997:
https://github.com/AuthMe/AuthMeReloaded/blob/master/docs/hash_algorithms.md

argon2
https://github.com/p-h-c/phc-winner-argon2

scrypt
https://github.com/Tarsnap/scrypt

other
https://github.com/lenis0012/LoginSecurity-2/tree/master/src/main/java/com/lenis0012/bukkit/loginsecurity/hashing/active

table

![table](https://cloud.githubusercontent.com/assets/5884000/24369889/c2263606-1325-11e7-9937-ada14684a543.png)

ljacqu:
we can do it with https://github.com/wg/scrypt, which is even on Maven central

Eyremba:
This is useless. I already posted this some time ago:

> Maybe take a look at Argon2? Argon2 is currently the most secure password hashing algorithm, even more secure than scrypt. It could be a milestone too for 5.4.

> Original Argon2 source: https://github.com/P-H-C/phc-winner-argon2
Java implementation of Argon2: https://github.com/phxql/argon2-jvm

Argon2 is much more secure than Scrypt and much better. Implementing scrypt would just implement an old outdated algorithm...

timvisee:
Maybe, implementing some dynamic hashing solution might be the best idea. With that, I mean a solution that supports many known hashing algorithms without requiring changes in the code regarding hashing, with minimal code change (for the initial implementation). Except for a little configuration change however. It would be awesome if the server administrator would have the ability to configure what algorithm is used using some form of algorithm identifier string, whether that'd be a `bcrypt`, `scrypt`, `Argon2` or `/dev/random`.

The problem is, that a single algorithm isn't ideal for all situations. Generally speaking, a hashing algorithm that takes longer (in time) to hash, is more secure against brute force attacks. However, that also means that the longer it takes to hash, the more resource expensive it is. And that for each hash calculation that needs to be done when a user enters a password through the login command. Some hashing algorithms even occupy all available CPU cores, which can cause great performance issues on servers running lots of users, or on low performance servers. I've heard that Facebook even uses dedicated hashing servers to minimize the performance impact for users that are logging in on their regular servers, although I don't know whether that is true.

[...]

PR #1165, opened 2017-04-13
Implement ARGON2 hash

sgdc3:
#1150

### DIFF
--- a/docs/hash_algorithms.md
+++ b/docs/hash_algorithms.md
@@ -1,5 +1,5 @@
 <!-- AUTO-GENERATED FILE! Do not edit this directly -->
-<!-- File auto-generated on Sat Mar 25 00:15:27 CET 2017. See docs/hashmethods/hash_algorithms.tpl.md -->
+<!-- File auto-generated on Fri Apr 14 01:40:05 CEST 2017. See docs/hashmethods/hash_algorithms.tpl.md -->
 
 ## Hash Algorithms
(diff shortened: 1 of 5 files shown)


sgdc3:
@ljacqu we are only missing the test class i think

ljacqu:
Looks good so far but it doesn't work yet ;) A hash will never be checked against its password successfully; you need to override `comparePassword` so it goes over the Argon2 class.

You can generate the test class for Argon2 by using `AbstractEncryptionMethodTest#generateTest`

sgdc3:
> You can generate the test class for Argon2 by using AbstractEncryptionMethodTest#generateTest

How? XD

ljacqu:
```java
     /** Generates a test class for a given encryption method. Simply create a test class and run the following code,
     * replacing {@code XXX} with the actual class:
     * <p>
     * <code>@org.junit.Test public void generate() { AbstractEncryptionMethodTest.generateTest(new XXX()); }</code>
     * <p>
     * The output is the entire test class. */
```

So just create an Argon2Test class, and in it have the code posted above and you're good to go ;)

sgdc3:
Test is failing, any idea?

ljacqu:
Hopefully fixed it. Reason is the test was confused because Argon2 extends `UnsaltedMethod` but it's very much salted&mdash;we just don't have to worry about it ourselves.

PR #1172, opened 2017-04-22
DataSource changes

ljacqu:
Some changes to the DataSource methods to make stuff more specific, and to also add yaw & pitch to last location. 
Would you mind having a small look @Gnat008 ? :)

### DIFF
--- a/src/main/java/fr/xephi/authme/datasource/SQLite.java
+++ b/src/main/java/fr/xephi/authme/datasource/SQLite.java
@@ -591,18 +591,18 @@ public class SQLite implements DataSource {
     }
 
     @Override
-    public List<PlayerAuth> getLoggedPlayers() {
-        List<PlayerAuth> auths = new ArrayList<>();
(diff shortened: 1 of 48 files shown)


src/main/java/fr/xephi/authme/datasource/SQLite.java
```diff
@@ -601,12 +601,45 @@ private PlayerAuth buildAuthFromResultSet(ResultSet row) throws SQLException {
...
 
         String ip = row.getString(col.IP);
         if (!ip.isEmpty()) {
             authBuilder.ip(ip);
         }
         return authBuilder.build();
     }
+
+
+    private static void close(Statement st) {
```
ljacqu: Thanks for the review. 
When I tested this locally the sqlite driver was implementing an older version of jdbc such that the `isClosed` method doesn't exist. So I added back the specific implementations. I kept the utils class, thinking Sqlite will be updated sometime, which will allow us to use the same methods. It's probably more sensible to revert it and to create a utils class when it makes sense, so I'll do that :)

Issue #1188, opened 2017-04-29
Stop creating EncryptionMethod objects excessively


ljacqu:
See PasswordSecurity -> each time a player wants to log in we emit an event and create an EncryptionMethod instance. In surely 95% of the cases it is fine if we just do this once for each `HashAlgorithm` setting. Introduce a configuration for this.

ljacqu:
on reload any EncryptionMethod needs to be cleared, though, regardless whether the hashalgorithm property changes or not

PR #1191, opened 2017-04-30
#1188 Create and keep encryption method instance

ljacqu:
- Avoid creating the same object over and over; instead keep it and only change it on settings reload


ljacqu:
Hi @sgdc3, this is my proposition for #1188 and it introduces a breaking change. I removed the player name from the encryption method event and we'll only throw it once on startup / reload. I think this breaking change is OK because the `EncryptionMethod` interface has the player name, so if someone is doing player name-related logic he just needs to wrap it _inside_ the encryption method he passes. Is this OK in your opinion?

sgdc3:
It looks ok, i think only a small number of people used such event, i don't think a small change would cause huge issues ;)

Issue #1090, opened 2017-02-11
Commands.yml new command event: onSessionLogin


HLF-Raideer:
Hello!

I want to suggest you a new command event to the commands.yml file.
Actually there is onLogin, onJoin, onRegister, my idea is to add a new **onSessionLogin** event to identify players who were already logged in. I want to execute commands if they are logged in during the login.

That's all! Thank you for maintaining this awesome plugin :)

molor:
another improvement: Make the ability to limit these events for some players only.
For example, **onLogin** event will be executed only in case if the player has a nickname **plaaayer**.

ljacqu:
@molor Thanks for your input. To keep some order I created a new issue with your suggestion. You (and everyone else) is very welcome to post more examples you would find useful :)

ljacqu:
@sgdc3 Would this be a good story for you? :) It would be nice if someone else worked on the forced commands to share the knowledge.

sgdc3:
@ljacqu It's fine for me, if you want i can work on this, i'm just a little busy this weekend ;)

ljacqu:
@sgdc3 It would be nice! :) And there's no hurry :smile:

ljacqu:
@sgdc3 Is this on your radar? ;) :smile:

sgdc3:
yep!

HLF-Raideer:
And when do you want to add it to the plugin? :D
.. because i posted this ticket 2 months ago xD

sgdc3:
When i'll have time, maybe today, maybe the next week, who knows :P

sgdc3:
Ok, i'll try to do that this weekend

sgdc3:
@HLF-Raideer Does it work? ;)

ljacqu:
It might be worth to note that these commands are run at a time when the player is not yet logged in.

sgdc3:
I think we can close that.

ljacqu:
tested successfully

PR #1198, opened 2017-05-04
Implement onSessionLogin trigger for commands.yml

sgdc3:
#1090 
Needs test methods, @ljacqu any advice?


ljacqu:
Cool! :smile:
On mobile right now – whatever the other tests do you can probably also do for this new section

sgdc3:
@ljacqu Could you explain me why onLogin has 2 tests? (the complete/incomplete commands.yml check)

ljacqu:
@sgd3 Unit tests should kind of cover the cases we expect to happen — one case is there being commands in the commands.yml, but another case is that there is an empty commands.yml. When I wrote these tests I was just trying to cover as many cases as possible with reasonable effort.

ljacqu:
hi @sgdc3, this is the last issue in the 5.3 milestone

sgdc3:
I'm atm busy with exams, i'll try to do that this weekend

PR #1202, opened 2017-05-05
Estonian translation of AuthmeReloaded

Rvcboy:


sgdc3:
Thank you ;)

ljacqu:
Thanks a lot! :) We've updated the file to sync with the latest message changes and got some new todo's --- https://github.com/AuthMe/AuthMeReloaded/blob/master/src/main/resources/messages/messages_et.yml in case you have time for a second round, it's much appreciated :)

PR #1214, opened 2017-05-12
Update messages_zhcn.yml

Playhi:
I think maybe the annotations also need to be translated.


sgdc3:
Idk, what do you think @ljacqu ?

ljacqu:
They don't _need_ to be translated but in a language file, nothing speaks against translating everything :)

ljacqu:
> Pull requests that have a failing status can’t be merged on a phone.

I can't merge atm :smile:

PR #1268, opened 2017-06-30
Updated messages_vn.yml and added help_vn.yml file

anhcraft:
Updated messages_vn.yml: translated time unit, etc
New translation for help_vn.yml file


Xephi:
Thanks for your contribution :)

PR #1283, opened 2017-07-13
Small fixes in Polish translation and addition of missing colors.

RikoDEV:
I could do it before, I'll do it now. 👍


sgdc3:
Thank you

Issue #1302, opened 2017-07-30
MSSQL Support


fmohican:
Its possible to provide support for Microsoft SQL Server ?
https://www.microsoft.com/en-us/sql-server/sql-server-2016
Thanks you.

Xephi:
Hi, what's the problem with Microsoft Sql Server? 

Since we already support SQL syntax, is it something into the sql engine that change from all other existing ?
Since we support MySQL, we also support PostgreSQL (without clustering optimisations), InnoDB, etc.. engines, i thinks it actually works well with Microsoft SQL Server, have you test it ?

sgdc3:
@Xephi i think the dialect is different

fmohican:
@Xephi nop i didn't test it, i will test in few moment.

fmohican:
isn't work
https://pastebin.com/phyTP0pM

Xephi:
Huum, ok, @sgdc3 HikariCP alreay handle the Microsoft SQL Server driver

```java
public DataSource getDataSource() {
     HikariConfig config = new HikariConfig();
     config.setDataSourceClassName("com.microsoft.sqlserver.jdbc.SQLServerDataSource");
     config.addDataSourceProperty("serverName", serverName);
     config.addDataSourceProperty("port", port);
     config.addDataSourceProperty("databaseName", databaseName);
     config.addDataSourceProperty("user", user);
     config.addDataSourceProperty("password", password);

     return new HikariDataSource(config);  //pass in HikariConfig to HikariDataSource
}
```
We can add the handle easily i think

fmohican:
would be nice, on hosted server i have to install mysql just for minecraft.... rest of application use MSSQL. xD

sgdc3:
@Xephi, no we have to add the driver, create a MsSql class and change the sql commands

sgdc3:
@fmohican could you try this jar please? https://2079-8662641-gh.circle-artifacts.com/0/tmp/circle-artifacts.icdOyru/AuthMe-5.4-SNAPSHOT.jar

(in the config set the database type as MSSQL before starting)

sgdc3:
try to register/login/changepassword/logout etc
if everything works fine we are almost ready to merge it into the master branch

fmohican:
sure I will test today and I'll get back with reply.

fmohican:
isn't work, i tryied with all port still get same error.
Default port of MSSQL is 1433 or 1434 TCP
[...]

PR #1303, opened 2017-07-30
First attempt to implement MsSQL datasource

sgdc3:


sgdc3:
#1302

sgdc3:
lol codeclimate isn't happy

ljacqu:
Sorry for the late reply & thanks for waiting. I checked out the branch and went through the changes. A+ :smile:

sgdc3:
good, but it doesn't work xD

sgdc3:
Too difficult all the queries are different, i think PostgreSQL support would be better.

sgdc3:
I'll keep the branch as it contains the SQLDataSource abstraction class that might be useful in the future

PR #1331, opened 2017-09-16
Login performance

sgdc3:


sgdc3:
@ljacqu what do you think about this? ;)

ljacqu:
hi @sgdc3, this looks cool! This will make the verifications happen in async for server implementations where it's possible. Hopefully this is helpful for us. One thing I'm unsure about is if `PermissionsManager#hasPermissionOffline` will have differences in behavior vs. the currently used `PermissionsManager#hasPermission`. That might be worth to have a quick look at.

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -190,46 +192,73 @@ public void onPlayerJoin(PlayerJoinEvent event) {
...
+        onJoinVerifier.checkIsValidName(name);
+
+        // Get the auth later as this may cause the single session check to fail
+        // Slow stuff
+        final PlayerAuth auth = dataSource.getAuth(name);
+        final boolean isAuthAvailable = auth != null;
+        onJoinVerifier.checkKickNonRegistered(isAuthAvailable);
+        onJoinVerifier.checkAntibot(name, isAuthAvailable);
+        onJoinVerifier.checkNameCasing(name, auth);
+        onJoinVerifier.checkPlayerCountry(name, ip, isAuthAvailable);
```
ljacqu: One iteration beyond this (but only once we resolve this permission behavior change!) is to look at whether we really need the auth object at all? As far as I see here from a quick look, we only want to know whether `isAuthAvailable` and then I only see it used in `checkNameCasing(name, auth)`.
If instead we can have a dedicated method to return the real name (similar to `DataSourceResult<String> getEmail(String)`) we avoid constructing PlayerAuth objects, which is more costly.

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -190,46 +192,73 @@ public void onPlayerJoin(PlayerJoinEvent event) {
...
+    }
+
     // Note #831: AsyncPlayerPreLoginEvent is not fired by all servers in offline mode
     // e.g. CraftBukkit does not fire it. So we need to run crucial things with PlayerLoginEvent.
     // Single session feature can be implemented for Spigot and CraftBukkit by canceling a kick
     // event caused by "logged in from another location". The nicer way, but only for Spigot, would be
     // to check in the AsyncPlayerPreLoginEvent. To support all servers, we use the less nice way.
 
+    private boolean isAsyncPlayerPreLoginEventCalled = false;
```
ljacqu: Fields go to the top of the class :) Thanks

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -190,46 +192,73 @@ public void onPlayerJoin(PlayerJoinEvent event) {
...
+
+    @EventHandler(priority = EventPriority.HIGHEST)
+    public void onAsyncPlayerPreLoginEvent(AsyncPlayerPreLoginEvent event) {
+        isAsyncPlayerPreLoginEventCalled = true;
+
+        final String name = event.getName();
+
+        if (validationService.isUnrestricted(name)) {
+            return;
+        }
```
ljacqu: Here, like in the non-async method, would it be worth to check `event#result`? I don't think it's dramatic as it is right now, but we'd be more consistent with the same check.

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -190,46 +192,73 @@ public void onPlayerJoin(PlayerJoinEvent event) {
...
-            onJoinVerifier.checkNameCasing(player, auth);
-            onJoinVerifier.checkPlayerCountry(player, event.getAddress().getHostAddress(), isAuthAvailable);
-        } catch (FailedVerificationException e) {
-            event.setKickMessage(m.retrieveSingle(e.getReason(), e.getArgs()));
-            event.setResult(PlayerLoginEvent.Result.KICK_OTHER);
-            return;
-        }
-
         teleportationService.teleportOnJoin(player);
+
```
ljacqu: This introduces a change in behavior I'm not sure is intentional—previously with the red block we did all of the onJoin verifications before and kicked the player. With the green addition the player gets teleported before any checks are made.
This is maybe better because this way a player is teleported as fast as possible (since we don't want to show the original location before login in many cases) ? But at the same time in the case of a bot attack we might be teleporting a lot of players before determining that they have no right being online. Difficult...

src/main/java/fr/xephi/authme/listener/PlayerListener.java
sgdc3: Forgot to move that, sorry xD

src/main/java/fr/xephi/authme/listener/PlayerListener.java
sgdc3: I think it would break some features like the single session kick message.

src/main/java/fr/xephi/authme/listener/PlayerListener.java
sgdc3: Teleporting players to a same spot shouldn't be resource intensive

sgdc3:
> Edit: Actually I think it's quite a difference given that if we don't hook into a specific perms system, the default permission (true / op / false) will be considered. IMHO a breaking change I'm not sure is worth to introduce?

Could we talk about that on discord? @ljacqu

src/main/java/fr/xephi/authme/listener/PlayerListener.java
ljacqu: Agreed ✔️

sgdc3:
Ready to merge?

ljacqu:
I'll test this locally and merge afterwards.

Issue #899, opened 2016-08-13
Email confirmation or release accounts registered by mail and have not logged in


hsombini:
Hello guys, have you ever considered a register system that player must confirm his account by doing something or the account would be deleted so other player can register it?

Maybe just send the password through mail like the plugin already do, but if the player doesn't login in X amount of hours, the account will be deleted.

I feel that registering by email, many accounts would be will be registered but the owner will never login, maybe a typo on email or something else, with that feature we can make sure that the accounts will get released fast.

ljacqu:
Let's try not to shove everything in 5.2 please

ColorsASD:
I would be happy for that too.

Right now it looks like this to me:

I have the following options after entering an incorrect Email or a fake Email:
1.: Or you will never restore your account to.
2.: Or you just can’t play because you sent the password to an email you don’t have.

There are a lot of problems with users, and since identities can only be verified with an email address, I’m amazed that this improvement hasn’t happened yet.

The problem is that there is no way the player can be identified. And so anyone could come in for forgetting their password.

Someone has spent real money, and someone just wants to break into your account.

So I would ask for at least one feature that if you enter an email, it will need to be verified and only then can registration continue. (It’s almost everywhere and so it’s logical I don’t understand why they don’t think about it here. This is a huge security issue. Especially on a server where hundreds play.)

😄🙄

Issue #814, opened 2016-07-02
[Dev] Review datasource column defaults


ljacqu:
- Does it really make sense to have x, y, z last positions default to 0? If no data is available player should probably not be teleported at all in a lastposition context. Same for last world.
- realname in SQLite defaults to Player but not in MySQL...? But in MySQL it is still defined `NOT NULL`
- email defaults to `your@email.com` - now we essentially check against this value instead of an intuitive NULL.

sgdc3:
also the 127.0 0.1 default ip xd

ljacqu:
I propose to close this even though I've left the location columns untouched. The big learning during this issue and #792 is that we kind of need NOT NULL columns to have a default value - SQLite requires that a NOT NULL column have a default value if it is added to a table later on, and various forum software we want to support has requirements for e.g. the email to be not null. Here the default value comes into play.

On PlayerAuth level we can handle these default values by replacing them with NULL. So if you get a PlayerAuth from the data source and the email returned from the database was `your@email.com`, then `playerAuth.getEmail()` will return `null`. Of course, if we interact differently with the DB (e.g. by fetching the email only), then these values will of course NOT be replaced to null. 

So despite the changes it seems that we're stuck with the default values. The best way to handle it is to be aware of them and have some constant / a method available to check for this, as we have in `PlayerAuth` now or e.g. `Utils#isEmailEmpty`.

Issue #792, opened 2016-06-24
Store registration IP and date


Twonox:
improve:   _/authme getip <pseudo>_  -  Display the first IP of the player (when registered) and old IPs.
add:  _/authme check <pseudo>_  -  Show the date of registering and show if player is registered.
add:  _AntiVPN_  -  The contry blacklisting isn't 100% efficiency, also why not add an AntiVPN like https://www.spigotmc.org/resources/protect-your-server-with-the-new-anti-vpn-plugin-1-7-and-1-8.18726/ This plugins is very efficiency but isn't up to date, if you add this to Authme with improvement, it can be better.

sgdc3:
Maybe on future:
1) we only store the last ip in our database
2) we don't store the registration date
WontFix:
3) we are not an antibot plugin xD

Twonox:
1) You can't store it?
2) You can't store it?
3) LOL, then what is that? xD
```
Features:​
Built-in AntiBot System!
```

ljacqu:
@Twonox for 1 and 2: we could store it, of course. But this complicates the application by having two more fields. It's not just two more columns in the database, but it's two additional properties to make them configurable, additional error handling to make sure the columns are there and to create them if not, additional logic to retrieve these column values from our various data sources.
And that's fine, it's nothing too complex, but this will all be additional code that we will then have to maintain in our application as the application will change. It essentially boils down to this: is the additional code worth the benefit it brings? If more people would like this, I'm up for it. Otherwise, no.

3: Our antibot system is fired when too many unregistered accounts join in a configurable interval, kind of like a "bot invasion". Then it kicks in and prevents more from logging on. This has nothing to do with detecting proxies or VPN connections.

ljacqu:
Granted, if someone else would want to contribute to AuthMe, 1 and 2 would make for a great introductory story to get to know the project.

sgdc3:
Yeah, you're right ;)

ljacqu:
This creates changes in the DB and in the semantics for purging, so although I'm working on this I'm moving this to 5.4 because I don't want to rush it.

sgdc3:
@ljacqu Could we close this?

ljacqu:
Yes, this is done—finally

PR #1332, opened 2017-09-17
Registration date and IP // null columns in DB

ljacqu:
Still a work in progress—PR for #792 and #814. I open up a PR this early for discussion and a preliminary review if someone has the time.

Work so far
- added columns for registration date and IP
- change MySQL to support not null for lastlogin and last_ip

Open points
- SQLite migration
- Need to handle purging logic... because after the change we will have accounts with `null` registration date, but also accounts with a `null` lastlogin date. Not sure what behavior is desired...
- I want to remove the default value for realname and email while we're doing changes to the DB (#814)

Any comments about the code so far or any ideas how the purging should be changed? Our current purging feature is basically to purge inactive accounts, which by two fields would basically be something like `purge if max(lastLoginDate, registrationDate) > NOW() - 40d` IMHO (if `40d` are configured)
Not to confuse with _Release accounts registered by mail and have not logged in_ #899.

Also what about max registration per IP? Do we go by registration IP or last IP?


games647:
> purge if max(lastLoginDate, registrationDate) > NOW() - 40d

I think this is the best solution.

>Also what about max registration per IP? Do we go by registration IP or last IP??

I think it should be last IP and if last IP is null, because the player not logged in yet, we could take registration IP. This should consider:
* Player joined the first time and registered more accounts on the same day (registration IP == last IP)
* Existing player joined the server, but created another account later the day (last IP == current IP)
* Player registered on the forums without login, but then registered another account by directly joining the server without forum registration in case that's possible (registration IP == current IP)

HexelDev:
> Also what about max registration per IP? Do we go by registration IP or last IP?

I'm going to define the two cases to analyze them better
###### 1st: registration IP

>  This case will use a static reference to the saved IPs list, **'x' accounts every IP**.

###### 2nd: last IP

> The dynamic nature of the IPs will interfere with this field, it will deny more then **'x' conteporary connections using the same IP**.

We could go with the first one to have some static point of interaction, the second one is not the scope of this function, am I right?

HexelDev:
@games647 you just described the complex system I was thinking about... 
But I was not sure if it could really be worth it, a system based on dynamic data that could be analyzed only during specified time windows based on matching rules 

(I really like the idea but is it worth it? I can't find the answer XD)

games647:
@HexelDev I rather thinks it's not worth it, because you can workaround it this limitation in many ways like disconnect your router or wait for the next day without logging in with your existing account.

HexelDev:
Yeah that's my same point...

It's not possible to create a system that stem these cases... It should be too complex and still have weak points :/

HexelDev:
> Need to handle purging logic... because after the change we will have accounts with null registration date, but also accounts with a null lastlogin date. Not sure what behavior is desired...
We could set all the `null` date value to the current date (`NOW()`), this will let the purge system works without issues.

ljacqu:
That sounds great @HexelDev, without that default we'd have `reg_date` that can have NULL entries and also the `last_login` column, which would give us four cases to discuss:

```
name | reg_date | last_date
---- | -------- | ----------
 a   |  NULL    |  12
 b   |  13      |  NULL
 c   |  12      |  13
 d   |  NULL    |  NULL
```
- a) is what will be in the DB of anyone using AuthMe after this migration. Maybe also with some forum integration that is wrongly set up?
- b) is what might happen in the future with email registration: user registered but has never logged in.
- c) is what will be in the DB once a new user registers and logs in
- d) should never happen during gameplay, but converting data from another plugin might generate this.

I was going to argue against using a default value for registration date since with this PR we're actually trying to _remove_ default values—however, a default would prevent us from having case d), which would complicate things for us and for the user. In the end I'd agree with the `NOW()` default, because in regards to purging it's a good default that will never cause any record to be purged to soon.

ljacqu:
In order to actually finish this PR sometime I will have to revert the removal of default values for email and IP. While testing forum integrations I realized that this breaks our integration into certain forums that have `NOT NULL` on their email column. 

I have ideas how to tackle it, but since it provides no value to the user (besides having a slimmer database) I'll focus on #792 for the time being, as other issues depend upon these columns being added.

ljacqu:
Decision: commits so far get squashed and pusehd to a new branch, and the changes of the new branch will be:
- Add column for registration date and registration IP.
  - registration date is `NOT NULL` (as explained by @HexelDev in https://github.com/AuthMe/AuthMeReloaded/pull/1332#issuecomment-334418187)
    - reasoning: need this data for purging. Current date is a good default if we don't have the information.
    - the default is implemented ON APPLICATION LAYER -> no `DEFAULT` set over SQL
  - registration ip is nullable. If we don't have the info we don't have it. Period.
- Despite my previous comment, also change last IP and last date to be nullable
  - Signals that user has never logged on... Again, if we don't have the info, we don't have it and deal with it. The `DEFAULT` values in SQL must die.

PR #1342, opened 2017-10-01
Update messages_de.yml

Platinteufel:


ljacqu:
Good catch :) Shall we merge or is there still something open?

Platinteufel:
Just found some other points, should be ready to merge now. :)

ljacqu:
Tipptopp, danke für dini Änderige :) Demfall tuehni jetzt merge~

PR #1351, opened 2017-10-09
Introduce hasSession field in datasource

sgdc3:
That makes isLogged more consistent as it will be '1' only when the player is online.

TODO:
- Unit testing?


sgdc3:
What do you think? My main concerns are performance and security.

docs/config.md
```diff
@@ -40,6 +40,8 @@ DataSource:
     mySQLColumnEmail: 'email'
     # Column for storing if a player is logged in or not
     mySQLColumnLogged: 'isLogged'
+    # Column for storing if a player has a valid session or not
+    mySQLColumnHasSession: 'isLogged'
```
ljacqu: Intentional to have the same name as `mySQLColumnLogged`? Maybe have an explanation for that?

src/main/java/fr/xephi/authme/command/executable/authme/RegisterAdminCommand.java
```diff
@@ -51,38 +51,31 @@ public void executeCommand(final CommandSender sender, List<String> arguments) {
...
-                    commonService.send(sender, MessageKey.ERROR);
-                    return;
-                }
-                dataSource.setUnlogged(playerNameLowerCase);
+            if (!dataSource.saveAuth(auth)) {
+                commonService.send(sender, MessageKey.ERROR);
+                return;
+            }
+            dataSource.setUnlogged(playerNameLowerCase);
+            dataSource.revokeSession(playerNameLowerCase);
```
ljacqu: are the above two calls necessary as a player gets saved? I'd expect those are the defaults.

src/main/java/fr/xephi/authme/datasource/MySQL.java
```diff
@@ -566,6 +566,44 @@ public void setUnlogged(String user) {
     }
```
ljacqu: Initialization of `hasSession` if missing should be added to `checkTablesAndColumns()`

src/main/java/fr/xephi/authme/datasource/SQLite.java
```diff
@@ -467,6 +467,48 @@ public void setUnlogged(String user) {
     }
 
     @Override
+    public boolean hasSession(String user) {
+        String sql = "SELECT * FROM " + tableName + " WHERE LOWER(" + col.NAME + ")=?;";
```
ljacqu: `SELECT col.HAS_SESSION` would be better as it retrieves only what we want from the DB.

src/main/java/fr/xephi/authme/datasource/SQLite.java
```diff
@@ -467,6 +467,48 @@ public void setUnlogged(String user) {
     }
```
ljacqu: Like in the MySQL class, here I'm missing the initialization of the `hasSession` column

src/main/java/fr/xephi/authme/datasource/SQLite.java
```diff
@@ -467,6 +467,48 @@ public void setUnlogged(String user) {
...
+            pst.executeUpdate();
+        } catch (SQLException ex) {
+            logSqlException(ex);
+        }
+    }
+
+    @Override
+    public void revokeSession(String user) {
+        String sql = "UPDATE " + tableName + " SET " + col.HAS_SESSION + "=? WHERE LOWER(" + col.NAME + ")=?;";
+        if (user != null) {
```
ljacqu: Why check here that `user != null` if this is not done in MySQL or in `SQLite#grantSession`? I don't think it's necessary.

src/main/java/fr/xephi/authme/process/join/AsynchronousJoin.java
```diff
@@ -179,8 +179,9 @@ private void processJoinSync(Player player, boolean isAuthAvailable, Location lo
 
     private boolean canResumeSession(Player player) {
```
ljacqu: What would you think about introducing a `SessionService` that would encapsulate the session handling logic? This has become a large method that I don't really feel is a good fit inside `AsynchronousJoin`.

src/main/java/fr/xephi/authme/process/logout/AsynchronousLogout.java
```diff
@@ -53,6 +53,7 @@ public void logout(Player player) {
 
         playerCache.removePlayer(name);
         database.setUnlogged(name);
+        database.revokeSession(name);
```
ljacqu: It looks like `revokeSession` and `setUnlogged` are always called together. If we merge this into one call we would save some DB operations, which would be good.

src/main/java/fr/xephi/authme/process/quit/AsynchronousQuit.java
```diff
@@ -82,8 +82,11 @@ public void processQuit(Player player) {
         playerCache.removePlayer(name);
 
         //always update the database when the player quit the game (if sessions are disabled)
-        if(!wasLoggedIn || !service.getProperty(PluginSettings.SESSIONS_ENABLED)) {
+        if(wasLoggedIn) {
             database.setUnlogged(name);
+            if(!service.getProperty(PluginSettings.SESSIONS_ENABLED)) {
```
ljacqu: here and above, would be nice to have a space before reserved keywords like `if` `while` `for` etc.
Same for the `AsynchronousLogin` addition

src/main/java/fr/xephi/authme/process/join/AsynchronousJoin.java
```diff
@@ -179,8 +179,9 @@ private void processJoinSync(Player player, boolean isAuthAvailable, Location lo
...
         final String name = player.getName();
-        if (database.isLogged(name)) {
+        if (database.hasSession(name)) {
             database.setUnlogged(name);
+            database.revokeSession(name);
             playerCache.removePlayer(name);
             if(service.getProperty(PluginSettings.SESSIONS_ENABLED)) {
```
ljacqu: `playerCache.removePlayer(name);` necessary? A player should never be in the PlayerCache at this point in time.

docs/config.md
sgdc3: LOL, i shoudn't code on the train.

src/main/java/fr/xephi/authme/command/executable/authme/RegisterAdminCommand.java
sgdc3: Right, good point

src/main/java/fr/xephi/authme/datasource/MySQL.java
sgdc3: Yeah, again, i shouldn't code when i'm on a crowded train xD

src/main/java/fr/xephi/authme/datasource/SQLite.java
sgdc3: Yeah, i left the previous code, i'll fix that asap

sgdc3:
isLogged will be just a flag that can be used by website integrations

src/main/java/fr/xephi/authme/process/join/AsynchronousJoin.java
sgdc3: Probably some code left there from the previous bungeecord cache implementation

src/main/java/fr/xephi/authme/process/logout/AsynchronousLogout.java
sgdc3: Nope, we don't do that when the player quit the server.

ljacqu:
Hi @sgdc3, if there are no changes on your end and you're satisfied with my additions, I invite you to merge the PR. ;)

ljacqu:
I took the liberty of squashing this PR so that we could have the issue number in the commit. This way, should we use Git blame months down the road to understand something, we can easily look up the issue for context.

sgdc3:
@ljacqu Uh, sorry yesterday i wasn't at home and i didn't read your comment ;)

PR #1371, opened 2017-10-21
#792 Force migration of SQLite when necessary

ljacqu:
As discussed on Discord, this changes the behavior of SQLite on startup to force a migration when it is necessary. Workflow:

- Detect if a migration is necessary
- Create a backup
- Perform the migration


sgdc3:
👍

ljacqu:
Thanks @sgdc3 :)

PR #1374, opened 2017-10-22
#1150 - Add Argon2 support

### LINKED ISSUE #1150: Add hashing support for Argon2

Xephi:
Wiki page : https://github.com/AuthMe/AuthMeReloaded/wiki/Use-Argon2-Password-HashMethod-with-AuthMe

Issue : #1150

### DIFF
--- a/src/test/java/fr/xephi/authme/security/HashAlgorithmIntegrationTest.java
+++ b/src/test/java/fr/xephi/authme/security/HashAlgorithmIntegrationTest.java
@@ -61,6 +61,14 @@ public class HashAlgorithmIntegrationTest {
         // given / when / then
         for (HashAlgorithm algorithm : HashAlgorithm.values()) {
             if (!HashAlgorithm.CUSTOM.equals(algorithm) && !HashAlgorithm.PLAINTEXT.equals(algorithm)) {
+                if (HashAlgorithm.ARGON2.equals(algorithm)) {
+                    try {
+                        System.loadLibrary("argon2");
+                    } catch (UnsatisfiedLinkError e) {
+                        System.out.println("[WARNING] Cannot find argon2 library, skipping test suite");
+                        continue;
+                    }
+                }
                 EncryptionMethod method = injector.createIfHasDependencies(algorithm.getClazz());
                 if (method == null) {
                     fail("Could not create '" + algorithm.getClazz() + "' - forgot to provide some class?");
@@ -99,4 +107,5 @@ public class HashAlgorithmIntegrationTest {
             fail("Found inconsistencies:\n" + String.join("\n", failedEntries));
         }
     }
+
 }
--- a/src/test/java/fr/xephi/authme/security/HashAlgorithmIntegrationTest.java
+++ b/src/test/java/fr/xephi/authme/security/HashAlgorithmIntegrationTest.java
@@ -2,6 +2,7 @@ package fr.xephi.authme.security;
 
 import ch.jalu.injector.Injector;
 import ch.jalu.injector.InjectorBuilder;
+import fr.xephi.authme.security.crypts.Argon2;
 import fr.xephi.authme.security.crypts.EncryptionMethod;
 import fr.xephi.authme.security.crypts.HashedPassword;
 import fr.xephi.authme.security.crypts.description.Recommendation;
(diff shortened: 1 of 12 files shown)


ljacqu:
Cool page!
- I don't think the config.yml excerpt on the page is correct. The hash setting is called something else
- How will we build the project and ensure that Argon2 is left out of the JAR?

Rest looks great

Xephi:
At my opinion, we already check if the library is present when we load the plugin ( for users)

For developpers: the only way is to check if the library exist on system when we do a test, i've to implement that check but idk where : 
```java
        try {
            System.loadLibrary("argon2");
        } catch (UnsatisfiedLinkError e) {
            ConsoleLogger.logException("Cannot find libargon2", e);
        }
```

Xephi:
Test fixed
Branch rebased on master and squashed onto one commit

src/test/java/fr/xephi/authme/security/HashAlgorithmIntegrationTest.java
```diff
@@ -61,6 +61,14 @@ public void shouldBeAbleToInstantiateEncryptionAlgorithms() {
         // given / when / then
         for (HashAlgorithm algorithm : HashAlgorithm.values()) {
             if (!HashAlgorithm.CUSTOM.equals(algorithm) && !HashAlgorithm.PLAINTEXT.equals(algorithm)) {
+                if (HashAlgorithm.ARGON2.equals(algorithm)) {
+                    try {
+                        System.loadLibrary("argon2");
+                    } catch (UnsatisfiedLinkError e) {
+                        System.out.println("[WARNING] Cannot find argon2 library, skipping test suite");
+                        return;
```
games647: Why not `continue` here?

Xephi:
Oh shit, well catched

ljacqu:
I have to admit I dislike the entire handling of Argon2 because of the additional steps it requires. I was silently hoping that the plain-java library someone mentioned in the issue would mature and be put on Maven Central

ljacqu:
Having said that, does anything speak against a static method on the Argon2 class to check if the library is loaded? That would be a potential place to have your check as a method in the code :)

ljacqu:
Hi @Xephi, if you allow me to push to this branch I'd like to move the Argon2 library check down from the abstract class to the Argon2Test class itself

Xephi:
You can write on that branch for sure, i think you already got rights x) Do the changes you that you think are better

ljacqu:
I had the rights but didn't want to interfere in case you were working on something ;) 
I did a push, thank you. With my commit the check for the library is done via one method in Argon2, and I've added the link to your Wiki page in the error messages. For that I had to shorten the Wiki link a little.
For me this is now fine, if you have no further changes I invite you to merge it (_Squash and merge_ so that we'll have one commit under your name :)). Cheers!

Eyremba:
> I have to admit I dislike the entire handling of Argon2 because of the additional steps it requires. I was silently hoping that the plain-java library someone mentioned in the issue would mature and be put on Maven Central

Yep, I think you mean this one: https://github.com/andreas1327250/argon2-java - mentioned in #1150 

But why don't we use the plain-java version by andreas1327250? :P

sgdc3:
>But why don't we use the plain-java version by andreas1327250? :P

> Fair warning: this code is not well tested and not reviewed. It is not ready for production, use it on your own risk. Same applies for underlying Blake2b implementation.

Eyremba:
> Read the project readme:

Dude, nearly EVERY private developer uses such disclaimers in their projects to prevent that they get sued when a big company uses their code and has problems with it.
If you test the code by yourself, you will notice that it is more than suitable for AuthMe.
This here is by the way Minecraft, and not your banking account.^^
So it should be more than sufficient.

sgdc3:
@ljacqu what do you think?

Xephi:
I can do a new branch with Java Implementation for tests only, actually keep our own implementation for production, and change for thus when it will be more stable
But we have to hardly test the implementation before put this in production

ljacqu:
Late to the party, but:

> This here is by the way Minecraft, and not your banking account.^^

is exactly why I don't understand why we added Argon2 in the first place, since it either requires a lot of special handling or using some shaky implementation ;)

Eyremba:
> is exactly why I don't understand why we added Argon2 in the first place, since it either requires a lot of special handling or using some shaky implementation ;)

Why do nearly all software developers encrypt data with AES?!
Right, because it is the standard for encryption.

And Argon2 is the new standard for password hashing and password verification. (Since Argon2 won the official password hashing competition.)

Issue #1311, opened 2017-08-10
Email verification code - hijacked accounts (using email)


HexelDev:
Adapted from [#1119](https://github.com/AuthMe/AuthMeReloaded/issues/1119#issuecomment-321056716) to manage it better.

@Maxetto 
> I'm providing also another security issue that relates to this.
> If you hack an Account who **has email**, you can unregister it and register it back in order to give it another email and make it impossible to recover.

HexelDev:
The idea:
When an user (that has an email associated with the account) execute the `/unregister` will be forced to insert a **confirmation captcha**.

This captcha will be sent via mail and it will confirm the account deletion.

Maxetto:
Or otherwise, after unregistration, if re-registration happens within a server restart, email is restored from old account.

ljacqu:
Some unfinished thoughts:
- Email captcha sounds lovely and is easy to implement. It should be fine to just store the codes in memory, i.e. we'd lose the codes once the server restarts.

- Captcha is required for changing the email address then, too, if we want to be fully safe.

- Should a server user be able to enable this? Our philosophy says yes—our config.yml however is creaming "stop bloating me"...

- An extended solution to this might relate to #681 if we introduce something like a "frozen" account state, which could be some state before an account gets deleted from the AuthMe database. For the linked issue it's interesting to avoid exploits; on here it's interesting as to provide a time window to the original user, should a malicious user unregister his account.

HexelDev:
I agree with the other ideas.

> An extended solution to this might relate to #681 if we introduce something like a "frozen" account state, which could be some state before an account gets deleted from the AuthMe database. For the linked issue it's interesting to avoid exploits; on here it's interesting as to provide a time window to the original user, should a malicious user unregister his account.

We could use the frozen state as an alternative to the captcha confirmations...

E.g.
We could provide 2 type of confirmation security:
- captcha confirmations
- 'frozen' state

While captcha confirmations is used:
- We can execute the deletion (and every other operation) after the confirmation.

While 'frozen' state is used:
- The operation will be confirmed after an 'x' configured time and can be reverted.

Because if the user email has been hijacked too, every step of security (like a 'frozen' state) will be useless... It will be an eternal battle between the user and the other guy to obtain the account...

HexelDev:
@ljacqu What do you think about the above ideas? :)

ljacqu:
[...]

PR #1382, opened 2017-10-27
Email verification code

HexelDev:
Email verification code.

Issue #1311


src/main/java/fr/xephi/authme/command/executable/verification/VerificationCommand.java
```diff
@@ -0,0 +1,59 @@
...
+    @Inject
+    private VerificationCodeManager codeManager;
+
+    @Override
+    public void runCommand(Player player, List<String> arguments) {
+        final String playerName = player.getName();
+
+        if (!codeManager.isEnabled()) {
+            ConsoleLogger.warning("Mail API is not set");
+            commonService.send(player, MessageKey.INCOMPLETE_EMAIL_SETTINGS);
```
ljacqu: This is not necessarily true. I'll take care of this (dropping a comment so I don't forget)

ljacqu:
@HexelDev, thanks for doing this! You did a great job at implementing this. This was a big feature to add with a lot of unclear details. You faced these changing assumptions / requirements head on and produced a solid implementation.

What I changed before merging was mainly moving the check for the permission within the verification code service so I can just call one method when I just care about answering "is a verification code required or no?". Besides that I removed the commented out image code and would vote not to implement that unless someone requests it.

Outlook:
- We should discuss about error handling. If an email cannot be sent (`EmailService#sendVerificationMail` returns a boolean indicating success) we do not handle that in the verification code service. This blocks a user from being able to do anything until emails can be sent.
- I think we could revisit the methods on `RecoveryCodeService`. From the naming, `isVerificationRequired` and `isCodeRequired` kind of sounds like the same to me. I also wonder if we shouldn't move more of the logic into that service instead of calling a handful of methods in `VerificationCommand` (I would expect to just call `verificationService#checkCode` or something).
- Could we add `/verify` also as label? Would be in line with our `/register`, `/login`, etc. commands

HexelDev:
> We should discuss about error handling. If an email cannot be sent (EmailService#sendVerificationMail returns a boolean indicating success) we do not handle that in the verification code service. This blocks a user from being able to do anything until emails can be sent.

We could add an error message that notify the user about this issue

> I think we could revisit the methods on RecoveryCodeService. From the naming, isVerificationRequired and isCodeRequired kind of sounds like the same to me. 

We could handle it better, that was my main "issue" during the coding process

- `isVerificationRequired` should define if the user is able to access the verification process
  - checks if the _emails are enabled_, the _user has a real email_, the _user has the permission_ and the _user has not been verified yet_.
- `isCodeRequired` should define if the user has a code and must use it to verify his identity
  - checks if the _user has a code_ and has not been verified yet.

> I also wonder if we shouldn't move more of the logic into that service instead of calling a handful of methods in VerificationCommand (I would expect to just call verificationService#checkCode or something).

I'm not a lot into the project, but as I saw in other commands (in a more in a more synthetic way) the messages where handled there, providing a good visual feedback, in my opinion, but obviously it's more a feedback than a smooth code...

> Could we add /verify also as label? Would be in line with our /register, /login, etc. commands

Yeah, that was the idea... I forgot to implement it ^.^

HexelDev:
@ljacqu Should we open an issue for this improvements?

ljacqu:
@HexelDev If you could, that would be great. I'm especially interest in the error handling.

HexelDev:
@ljacqu Ok, I'll create a new issue this evening

PR #1385, opened 2017-10-28
#792 Migrate last IP column if is NOT NULL without default value

ljacqu:
Last change before #792 can be verified / closed ;)

On MySQL some time before 5.2 the last IP column was initialized as `NOT NULL` but no default value was provided. Now that we don't set a "last IP" value upon registration the database is unhappy. This is the same migration as we already do it in SQLite, so it's good for both to behave similarly.

If the column is required to be NOT NULL (forum integrations), this can be switched again with `/authme debug mysqldef`, which will add the last IP column as NOT NULL incl. a default value.

Given the amount of "migration code" I decided to move all of the MySQL migrations into their own class. Didn't create a common type with the SQLiteMigrater because they don't have much in common but if you feel it would be sensible, we can talk about it. ;)

### DIFF
--- a/src/main/java/fr/xephi/authme/datasource/MySQL.java
+++ b/src/main/java/fr/xephi/authme/datasource/MySQL.java
@@ -20,7 +20,6 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.sql.Types;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


ljacqu:
Force-pushed to remove `@DataFolder` from `SqLiteMigrater`: not instantiated by injection so the annotation does nothing.

Eufranio:
Getting `[AuthMe] Error during SQL operation: [SQLException]: Field 'ip' doesn't have a default value`. Is that because of this?

ljacqu:
@Eufranio Exactly ;) Given Hexel's approval I'll merge this.

PR #1390, opened 2017-10-29
Fix Java 9 support for Pbkdf2Django

games647:
Java 9 removed the DatatypeConvert class and so I replaced it with the Base64 class which was introduced in Java 8. 

While working on that I discovered some issues with the Tests which were reproducible on [Travis](https://travis-ci.org/AuthMe/AuthMeReloaded/jobs/294205481). Suprisingly they occured only Java 9 which could be a bug in the Mockito implementation. 

1. AuthMeInitializationTest.initAuthMe NPE when accessing the mocked Logger [in PluginLogger of Bukkit](https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/plugin/PluginLogger.java#L26) 

Fix:
Change to anonymous logger
https://github.com/AuthMe/AuthMeReloaded/blob/d9b53f08b129ab8b1e8292d5999d3b4705d93b6e/src/test/java/fr/xephi/authme/AuthMeInitializationTest.java#L78

2.  SetPasswordCommandTest.shouldChangePassword NPE when trying to log something using ConsoleLogger

Fix:
Manually set the logger using TestHelper. 

3. RecoveryCodeServiceTest.shouldRemoveCode ClassCastException Integer cannot be cast to String 

Fix:
Change the generics to `<String, Integer>` instead of <String, String>. This seams pretty reasonable, because TimedCounter has <K, Integer>.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,9 @@ addons:
       - git
 
 language: java
-jdk: oraclejdk8
+jdk:
+  - oraclejdk8
+  - oraclejdk9
 
 before_script:
     - "sudo git clone https://www.github.com/P-H-C/phc-winner-argon2.git argon2-src"
--- a/pom.xml
+++ b/pom.xml
@@ -740,7 +740,7 @@
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
-            <version>2.11.0</version>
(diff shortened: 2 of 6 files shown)


ljacqu:
Awesome, thank you @games647!

Xephi:
wow, nice catches, that's a good PR :)

PR #1393, opened 2017-10-30
Implement bungee messaging, CacheDataSource sync across network

sgdc3:


HexelDev:
Codclimate 1 - @sgdc3 0

ljacqu:
@sgdc3 Am I right to assume you never actually ran this code?

sgdc3:
@ljacqu yes xD

ljacqu:
I spotted an exception ;) A unit test will uncover it

src/main/java/fr/xephi/authme/service/BungeeService.java
```diff
@@ -1,69 +1,191 @@
...
+    public void sendRefreshSession(String name) {
+        sendAuthMeBungeecordMessage(AuthMeBungeeMessageType.REFRESH_SESSION, name.toLowerCase());
+    }
+
+    public void sendRefreshQuitLoc(String name) {
+        sendAuthMeBungeecordMessage(AuthMeBungeeMessageType.REFRESH_QUITLOC, name.toLowerCase());
+    }
+
+    public void sendRefreshEmail(String name) {
+        sendAuthMeBungeecordMessage(AuthMeBungeeMessageType.REFRESH_EMAIL, name.toLowerCase());
```
ljacqu: Given that you're sending `name.toLowerCase()` to the same method from twelve different places, you could just use `toLowerCase()` inside `sendAuthMeBungeecordMessage`

src/main/java/fr/xephi/authme/service/BungeeService.java
```diff
@@ -1,69 +1,191 @@
...
-            messenger.registerOutgoingPluginChannel(plugin, "BungeeCord");
+
+        ByteArrayDataInput in = ByteStreams.newDataInput(data);
+        String subchannel = in.readUTF();
+        if(!"Authme".equals(subchannel)) {
+            return;
+        }
+
+        String type = in.readUTF();
+        switch (type) {
```
ljacqu: This switch doesn't handle all `AuthMeBungeeMessageType`.

src/main/java/fr/xephi/authme/service/BungeeService.java
```diff
@@ -1,69 +1,191 @@
...
+    private void handleRefresh(String name) {
+        if(!(dataSource instanceof CacheDataSource)) {
+            return;
+        }
+        CacheDataSource cacheDataSource = (CacheDataSource) dataSource;
+
+        if (cacheDataSource.getCachedAuths().getIfPresent(name) == null) {
+            return;
+        }
+        cacheDataSource.getCachedAuths().refresh(name);
```
ljacqu: I would like to have interface methods for `refresh` and `invalidate` so that we can stop performing checks à la `instanceof CacheDataSource`. If you add them as empty default methods to the interface the other implementations won't even need to be touched. :)

src/main/java/fr/xephi/authme/service/BungeeService.java
```diff
@@ -1,69 +1,191 @@
...
+        }
+
+        List<String> dataList = Arrays.asList(data);
+        dataList.add(0, "AuthMe");
+        dataList.add(1, type);
+        sendBungeecordMessage(dataList.toArray(new String[dataList.size()]));
+    }
+
+    public void sendLogin(String name) {
+        sendAuthMeBungeecordMessage(AuthMeBungeeMessageType.LOGIN, name.toLowerCase());
```
ljacqu: Up to you, just putting the idea out there: does it make sense to have dedicated methods for all of these cases if we could expose the `AuthMeBungeeMessageType` as an enum instead and have that coming in as an argument? You'd save on a lot of those specific methods and could easily add new entries to that enum without requiring a method counterpart to be added, too.

src/main/java/fr/xephi/authme/service/BungeeService.java
sgdc3: It's wanted, some messages are there for the bungee side plugin and future uses

src/main/java/fr/xephi/authme/service/BungeeService.java
sgdc3: Nice idea!

src/main/java/fr/xephi/authme/service/BungeeService.java
sgdc3: Right xD

src/main/java/fr/xephi/authme/service/BungeeService.java
sgdc3: That's because we might have some messages with extra arguments in the future, idk how to handle that correctly, any idea?

ljacqu: Personally, I'd handle that when the need for it arises. Like, you could have a `void sendMessage(AuthMeBungeeMessageType)` and later on the first time it comes up, you can easily overload it with a `void sendMessage(AuthMeBungeeMessageType, String... args)` method or whatever.

sgdc3:
I'll push the changes asap

sgdc3:
@ljacqu Test failing due to IllegalAccess? :/ Maybe some issue related to the default methods in the interface?

ljacqu:
I'll check it out this evening ^^

Issue #1350, opened 2017-10-08
AuthMe/LuckPerms server stall


Eufranio:
The server just stalled once a (maybe) random player joined. Stall-report: https://gist.github.com/Eufranio/dd8feaf2df8f1c82731aa80a32506a8c
Config: https://gist.github.com/Eufranio/6738caa3dd617e00d82f14b94ec31bdc

Running:
  AuthMe-5.4-b1808
  Thermos-1.7.10-1614-b58

sgdc3:
looks like a luckperms issue

Eufranio:
@lucko?

lucko:
@sgdc3 You're joining a future on the main thread.

This means the main thread is blocked, and pauses to wait for the future to complete. (not good)

You even commented this in the code....
https://github.com/AuthMe/AuthMeReloaded/blob/master/src/main/java/fr/xephi/authme/permission/handlers/LuckPermsHandler.java#L66

Don't see how you could possibly blame me, lol.

sgdc3:
Whoops, missed that point, sorry xD

sgdc3:
@lucko any way to set a timeout to the task?

sgdc3:
Uh right, ComparableFuture.get(timeout, TimeUnit) ;)

sgdc3:
@Eufranio Should have been fixed

lucko:
Blocking the main server thread for 1 second is still absolutely not ok.

You will absolutely kill the TPS, and create 1 second long lag spikes (worst case) every time you retrieve an offline user.

sgdc3:
> Also, it’s not a deadlock. You’re just waiting on a Future which can take a while to execute, as it does a lot of I/O, often with a remote database, and internal processing.

Probably a deadlock, because it took more than 30 seconds.

sgdc3:
@lucko How could we check the permission of an offline user witout blocking the main thread, it's impossible if we don't cache it somehow.

lucko:
Return a default value for the check, or actually perform the action which depends on the check asynchronously to start with.

sgdc3:
@ljacqu any idea?

ljacqu:
I'd like to forward the question to @lucko: when a player joins we do some permission checks (for VIP status and various restriction overriding permissions) based on the _name_ (not _Player_ object because it's not available in all joining events). I guess this falls into "offline player" check—how would you suggest we handle these?

ljacqu:
[...]

PR #1395, opened 2017-10-31
Add option to force using the Sync PlayerLoginEvent instead of AsyncPlayerPreLoginEvent

sgdc3:
#1350


sgdc3:
@Xephi no, to fix compatibility with some non vault compatible permission systems and luckperms. (At this stage the Player object isn't available yet, and the permission system doesn't contain it yet, via vault we can check permissions for offline players, but any non-vault compatible permission system or lukperms, due to its internal design, will be unable to resolve permissions when authme handles the OnJoinVerifications)

ljacqu:
Not that happy with the duplications this brings; could I submit an alternative implementation for this?

sgdc3:
Sure @ljacqu :)

PR #1402, opened 2017-11-06
Implement AuthMeBungee autologin

sgdc3:
There is a failing test due to cyclic dependency injection, @ljacqu could you take a quick look at this?

### DIFF
--- a/src/main/java/fr/xephi/authme/service/bungeecord/BungeeService.java
+++ b/src/main/java/fr/xephi/authme/service/bungeecord/BungeeService.java
@@ -1,46 +1,38 @@
 package fr.xephi.authme.service.bungeecord;
 
-import com.google.common.io.ByteArrayDataInput;
+import ch.jalu.injector.Injector;
 import com.google.common.io.ByteArrayDataOutput;
 import com.google.common.io.ByteStreams;
 import fr.xephi.authme.AuthMe;
-import fr.xephi.authme.ConsoleLogger;
-import fr.xephi.authme.datasource.DataSource;
 import fr.xephi.authme.initialization.SettingsDependent;
-import fr.xephi.authme.process.Management;
 import fr.xephi.authme.service.BukkitService;
 import fr.xephi.authme.settings.Settings;
 import fr.xephi.authme.settings.properties.HooksSettings;
 import org.bukkit.entity.Player;
 import org.bukkit.plugin.messaging.Messenger;
-import org.bukkit.plugin.messaging.PluginMessageListener;
 
 import javax.inject.Inject;
 
 /**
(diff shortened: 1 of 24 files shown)


ljacqu:
hey @sgdc3, you have a cyclic dependency. Your only choice is to split the BungeeService. The problem is that now you inject Management into BungeeService, but BungeeService gets injected into stuff that Management requires (e.g. AsyncQuit).

So in other words now the injector is like "to create BungeeService I need Management" -> "to create management I need asyncquit" -> "to create asyncquit I need BungeeService—uh oh"

src/main/java/fr/xephi/authme/service/bungeecord/BungeeService.java
```diff
@@ -113,9 +117,21 @@ public void onPluginMessageReceived(String channel, Player player, byte[] data)
             case MessageType.REFRESH:
                 dataSource.refreshCache(name);
                 break;
+            case MessageType.BUNGEE_LOGIN:
+                handleBungeeLogin(name);
```
ljacqu: missing a `break` here

HexelDev:
@ljacu that example is awesome ahhahaha

sgdc3:
>Your only choice is to split the BungeeService.

Any suggestion?

ljacqu:
> Any suggestion?

Sending and receiving. Inside the Management dependencies you're only sending out messages. The only annoying thing about that is the channel handling, which I guess you could keep in BungeeService maybe.
Without looking at it into too much detail I reckon you could do something like the following.

ljacqu:
img_7426

src/main/java/fr/xephi/authme/service/bungeecord/BungeeService.java
```diff
@@ -56,10 +48,14 @@ public void reload(Settings settings) {
             messenger.registerOutgoingPluginChannel(plugin, "BungeeCord");
         }
         if (!messenger.isIncomingChannelRegistered(plugin, "BungeeCord")) {
-            messenger.registerIncomingPluginChannel(plugin, "BungeeCord", this);
+            messenger.registerIncomingPluginChannel(plugin, "BungeeCord", injector.getSingleton(BungeeReceiver.class));
```
sgdc3: Is that correct? @ljacqu  :/

ljacqu:
I see the issue and don't have a solution :/ I need to check this out this evening on a PC

ljacqu:
I pushed my proposal. I haven't tested it in-game. Please make any necessary adjustments and kindly _squash_ this PR if you decide to merge it. We don't need six commits of back and forth in our Git history :)

sgdc3:
Looks good @ljacqu

Issue #1254, opened 2017-06-14
Command to see recently logged in players


ljacqu:
Relates to #1217:

As a server owner I want to be able to query a list of recently logged in players

PR #1429, opened 2017-11-28
#1254 Create command to see recently logged in players

ljacqu:
Impl for #1254 because I really want this

- Create datasource method to fetch most recent players by last login date
- Add command to view last logged in players


sgdc3:
:+1:

Issue #1046, opened 2016-12-31
New command event to Commands.yml: first login


belfedia:
I keep an eye on commands.yml

# Supported command events: onLogin, onJoin, onRegister
# onFirstLogin :
#  welcome:
#  command: 'give %p Cobblestone '
#  executor: 'PLAYER'**

Could you add a firstlogin event ? :)
I register only on my forum, and i search a clue to give some objects, make a global welcome to new player :)

ljacqu:
I'm guessing you want commands to always fire if a user is not registered—not only the first time a user appears on the server? :)

belfedia:
I want an event on first login, not first registrer or first time on server.
With inventory protect, i can give to new player item, only when they make a sucessfull /login.

SaitDev:
+1 for this

ljacqu:
hi @belfedia @SaitDev this is available on newer builds (n. 1880 and above) at http://ci.xephi.fr/job/AuthMeReloaded/
I would be thankful for some feedback - is something not how you expect it, or can we close this issue?

SaitDev:
its avaible on release 5.4 version too?

ljacqu:
It's not, it was added right after the 5.4 release

belfedia:
Thanks i try it :)

PR #1430, opened 2017-11-28
#1046 Add onFirstLogin to commands.yml

ljacqu:
- Allow to configure commands run on player's first login (login of player with a previously null lastlogin date)

Impl for #1046


sgdc3:
:+1:

PR #1446, opened 2017-12-07
Add new hash method

Stun3R:
This file is a new hash method which is used by a CMS : http://craftmywebsite.fr/
Now the owner of a Minecraft Server and Website based on this CMS could link his server with is WebSite and Database.

PR #1447, opened 2017-12-07
Update HashAlgorithm.java

Stun3R:
It's link to my last pull request. 
I recently add a new hash method in crypts/ directory. So this modification enable the class


Stun3R:
It's normal that the Travis CI Builde could not complete because the class that I added is in a pull request too... So it will work when the class AuthMeReloaded.src.main.java.fr.xephi.authme.security.crypts.MD5INSHA1.java will be added

Xephi:
J'ai bien reçu ton mail, normalement tu peux merge tes deux branches de patch en une seule et créer une unique pull request :)

games647:
Travis complains, because your code is in a different branch and so completely separated from this. The Java Compiler knows nothing about your other pull request and therefore cannot compile the project.  

It's like you would make those changes to two seperate project folders. Both folders contains a copy of the current master branch, but aren't connected together. (I hope this analogy makes it clear how git works)

You could just make another commit (changeset) to that other branch from pull request #1446. Then you change the same work space where your other changes are included.

PR #1461, opened 2017-12-27
Improve security of Console and Log files

Maxetto:
This is a security plugin and given it's purpose, having a simple option to show in plain text every password while instead there's no PLAINTEXT store option in the database (not anymore, as last time we discussed about this option there was still many refactoring to do), is stupid and unsafe.

Many people use this option more for bad than for good, probably no one will ever complain about this providing a good use case.

Latest step in security would be a way to encrypt or hide commands to command loggers, but this may be impossible with Bukkit API.

### DIFF
--- a/src/test/resources/fr/xephi/authme/settings/config-old.yml
+++ b/src/test/resources/fr/xephi/authme/settings/config-old.yml
@@ -315,8 +315,6 @@ Security:
     console:
         # Remove spam console
         noConsoleSpam: true
-        # Replace passwords in the console when player type a command like /login
-        removePassword: true
     captcha:
         # Player need to put a captcha when he fails too lot the password
         useCaptcha: false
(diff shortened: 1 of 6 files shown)


src/test/resources/fr/xephi/authme/settings/config-old.yml
```diff
@@ -315,8 +315,6 @@ Security:
     console:
         # Remove spam console
         noConsoleSpam: true
-        # Replace passwords in the console when player type a command like /login
-        removePassword: true
```
ljacqu: This should stay; it tests an old config.yml version

Maxetto: Ok, reverted that commit.

ljacqu: Cheers ;)

ljacqu:
Thanks for doing this :)

PR #1464, opened 2018-01-05
Captcha for register [unfinished]

ljacqu:
Starting to get really sick of this issue (#930)... Not my best work. :) Suggestions for improvements are very welcome.

**Still unfinished**, as messages on the limbo player need to be updated when a captcha was solved successfully.


ljacqu:
@sgdc3 raised an excellent point that the handling of circular dependencies will have to be improved from the current implementation. In order not to break the scope I'll keep with the current one within this PR and I'll create a technical follow-up issue to tackle that.

ljacqu:
I just found a way to resolve the circular dependency. Will probably merge this today

Issue #1467, opened 2018-01-06
Add additional level to message keys 


ljacqu:
As our YML message files are growing we need to ensure that the file stays maintainable. Currently we group messages with a YAML comment but this is troublesome when we manipulate YML files. It would be easiest if we changed the message keys to have their group as first level.

Currently we have:
```yml
# Login
usage_log: '&cUsage: /login <password>'
wrong_pwd: '&cWrong password!'
```

And after introducing a path in the key:
```yml
login:
    usage_log: '&cUsage: /login <password>'
    wrong_pwd: '&cWrong password!'
```

This would automatically solve #1087.

### Task
- Change message keys to new structure
- Implement a migration for old files
- Improve key names whenever possible (since we are changing the keys anyway)
- Change all placeholders to have the format `%placeholder%`

Twonox:
What do you mean with "Implement a migration for old files" ?

ljacqu:
Migration means that when AuthMe loads any of the existing message files, it should notice that it's in the old format and automatically migrate to the new format. 
The keys shouldn't be translated; that's how the message texts are identified by.

Twonox:
Okay I see, thanks ;)

ljacqu:
This is coming along quite well but it's a lot of work 😅

sgdc3:
:+1: :+1: :+1: :+1:

ljacqu:
Hi @sgdc3, would you like to verify if the behavior is OK? Any missing messages are added on the loading of a YML file from the local messages file (e.g. messages_fr.yml) and then whatever is left comes from messages_en.yml

sgdc3:
Do you need some testing or just an opinion?

ljacqu:
I have tested it, I'd just feel bad to close such a big issue without any kind of feedback :) So just an opinion essentially. :)

HexelDev:
Well done dude ;)

ljacqu:
@HexelDev Thank you :sweat_smile:

HexelDev:
<3

PR #1469, opened 2018-01-07
Update messages_fr.yml

Twonox:
@ljacqu 
Could you change the captcha variables (<theCaptcha> & THE_CAPTCHA) to %captcha% ?
Because there is a confusion with the argument messages.


ljacqu:
Thanks a lot, @Twonox! You're right; we should change all placeholders to be in the same format & easily recognizable. We can change them all to the `%placeholder%` format with #1467, since with that issue we'll be changing the message formats anyway.

Issue #1035, opened 2016-12-18
Move "command on x alts detected" setting to commands.yml


ljacqu:
```yml
        # Threshold of the other accounts command, a value less than 2 means disabled.
        otherAccountsCmdThreshold: 0
        # Command to run when a user has more accounts than the configured threshold.
        # Available variables: %playername%, %playerip%
        otherAccountsCmd: 'say The player %playername% with ip %playerip% has multiple accounts!'
```

This could be moved to the commands.yml file and would allow to define multiple cases when a command should be run.

- Create a class extending `Command` with an additional field to specify the number of alts required before the command is run
- Add setting to the commands.yml config files
- Add migration for moving the above setting to commands.yml

PR #1477, opened 2018-01-16
1035 handle alt accounts in commands yml

ljacqu:
First implementation part of #1035. What's still missing is the migration of the old settings to the new commands.yml structure. I'll do that outside of this PR.


sgdc3:
Looks good

ljacqu:
Thanks for your feedback. By removing that 5.2 migration of commands from config.yml to commands.yml we can get rid of a lot of code—cool! :)

PR #1492, opened 2018-02-01
#1467 Hierarchical message keys

ljacqu:
~This PR is not ready to merge yet — the tool task that adds verification comments still needs to be adjusted.~

Please check out the changes in MessageKey (and messages_en.yml as a result) and let me know if you have any suggestions or if you agree with the changes.

Besides introducing the new keys, this PR introduces a migrater class that has effect whenever a messages yml file is loaded. If any messages are missing within that yml file, it will be added based on the JAR's messages.


src/main/java/fr/xephi/authme/command/executable/authme/MessagesCommand.java
```diff
@@ -22,15 +16,6 @@
  */
 public class MessagesCommand implements ExecutableCommand {
```
ljacqu: I also need to update this command so that it doesn't take "help" as second argument anymore.

ljacqu:
This is ready to be merged, IMHO. Does someone want to look over the changes to the message keys?

ljacqu:
I just tested in-game and it's not ready to be merged yet: the migration replaces non-ASCII characters like é to question marks.

ljacqu:
OK, I'm launching another attempt at this! Everything should be done. I will test in-game tomorrow. 

If anyone wants to have a look at the code until then, please do! It would be much appreciated. This turned out to be quite a lot of changes :smile:

Issue #1497, opened 2018-02-03
Improve error logs when a yml config file cannot be loaded


ljacqu:
Inspired by #1495, we can improve the error messages that happen when an invalid YML file is loaded (messages / commands / help messages / config).

sgdc3:
yeah

PR #1506, opened 2018-02-23
#1497 Show specific message for invalid YAML files

ljacqu:
Impl for #1497.

- Wrap SnakeYAML exceptions when loading config.yml and commands.yml on startup into own exception type
- Handle exception type on startup with specific error message

### DIFF
--- a/src/main/java/fr/xephi/authme/service/yaml/YamlFileResourceProvider.java
+++ b/src/main/java/fr/xephi/authme/service/yaml/YamlFileResourceProvider.java
@@ -6,7 +6,7 @@ import org.yaml.snakeyaml.parser.ParserException;
 import java.io.File;
 
 /**
- * Extension of {@link YamlFileResource}, which throws a specific exception if a file cannot be loaded.
+ * Creates {@link YamlFileResource} objects.
  */
 public final class YamlFileResourceProvider {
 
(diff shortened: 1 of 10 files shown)


src/main/java/fr/xephi/authme/service/yaml/YamlFileResourceProvider.java
```diff
@@ -0,0 +1,30 @@
+package fr.xephi.authme.service.yaml;
+
+import ch.jalu.configme.resource.YamlFileResource;
+import org.yaml.snakeyaml.parser.ParserException;
+
+import java.io.File;
+
+/**
+ * Extension of {@link YamlFileResource}, which throws a specific exception if a file cannot be loaded.
```
ljacqu: Oops this is no longer an extension.
Initially I wanted to extend ConfigMe's YamlFileResource class but due to great design (cough, cough) it proved troublesome to wrap the initializing method there into its own exception (I don't have access to the file anymore, saving the file in an individual field also doesn't work because the exception will already have been thrown before the file field is set...).
I'll adjust the comment.

ljacqu:
Thank you! Being able to extend the YamlFileResource class would have been cooler because we could have handled the reload() case, too. I've noticed that a lot of classes of ConfigMe aren't really designed for extending, so in the next development iteration this will be addressed. :)

PR #1509, opened 2018-02-28
Be able to replace sender name in every message

### LINKED ISSUE #829: what variable as player name? (in language file)
as title said, what variable as player name in language file? i try %p, %player, %name, %username, all won't work.

Maxetto:
By setting a unique Tag (%username%), we can replace it with the sender's name inside the Send function.
This pull closes #829.

Some thoughts:

- It won't check for a Player instance, so even Console and Admin commands will replace it with the sender's name, if any. (I think Console gives an UUID or "ConsoleSender" as his name).
- Since Admin commands aren't yet translated, part of the above thought can be ignored for now.
- By setting it as a "special" tag, like %nl%, it is different than specific single message tags that only have 1 % at the beginning. This however may cause confusion (for example with the "%name" tag), we should add a comment line inside every translation stating which are the global replaceble tags and what they do.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -345,7 +345,7 @@
         <!-- Our Repo (Many libs) -->
         <repository>
             <id>codemc-repo</id>
-            <url>http://ci.codemc.org/plugin/repository/maven-public/</url>
+            <url>https://repo.codemc.org/repository/maven-public/</url>
         </repository>
 
         <!-- bStats Repo -->
--- a/src/main/java/fr/xephi/authme/message/Messages.java
+++ b/src/main/java/fr/xephi/authme/message/Messages.java
@@ -18,6 +18,8 @@ public class Messages {
     // Custom Authme tag replaced to new line
     private static final String NEWLINE_TAG = "%nl%";
 
+    private static final String PLAYER_TAG = "%username%";
+
     /** Contains the keys of the singular messages for time units. */
     private static final Map<TimeUnit, MessageKey> TIME_UNIT_SINGULARS = ImmutableMap.<TimeUnit, MessageKey>builder()
         .put(TimeUnit.SECONDS, MessageKey.SECOND)
@@ -51,7 +53,7 @@ public class Messages {
     public void send(CommandSender sender, MessageKey key) {
         String[] lines = retrieve(key);
         for (String line : lines) {
-            sender.sendMessage(line);
+            sender.sendMessage(line.replace(PLAYER_TAG, sender.getName()));
         }
     }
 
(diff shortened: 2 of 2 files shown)


Maxetto:
Add also a fix for failed builds

sgdc3:
Thank you

ljacqu:
Hi @Maxetto, a few remarks:

- The String#replace method you were using was fine and preferable over String#replaceAll. The latter uses regex so should only be used when you need to replace a pattern instead of a predefined text (it has a difference in performance).
- I think the tag replacement should have been performed in the `formatMessage` method: then you'd only perform the replacement at one place (and not in a loop) and it is performed at the same place as the `NEWLINE_TAG` replacement. 
   * I'm pretty sure that the replacement is not being done in texts retrieved via Messages#retrieve, so this would also fix that.
- If you like I can show you sometime how to merge branches in your IDE. Your commit "Fix build" could have been avoided by merging in the master branch which has the same changes.

Maxetto:
Hi, @ljacqu thank you for reviewing the code.
I've tried String#replace before but it didn't pass the tests because there was no "%username%" in the test strings and so it threw an NPE (https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#replace(java.lang.CharSequence,%20java.lang.CharSequence). I know replaceAll uses Regexes and I wanted to avoid it.

I've opted to go inside the send method because it already has the Sender parameter, instead of passing it to retrieve, then retrieveMessage and then formatMessage, I also know that it's nested inside a loop but as default messages are single line, the impact should be less heavy.

The goal of this was to be very simple and with little modification to the code base.

ljacqu:
In order to use String#replace you can choose one of two paths:
- Update the tests (as far as I know no CommandSender ever has `null` as name. We use it elsewhere with that assumptions and never had issues so far). The tests don't always define a name because it wasn't relevant until now. In general tests shouldn't dictate how code is written if it's just a matter of changing how something is set up. This would be my preferred approach since a test could be added or extended to test this new tag, too, which is a good idea to do when a new feature is introduced.
- Alternatively, before performing the replacement you could extract the name into a variable and provide empty string (or something else?) if it's null. 

Thankfully, Messages#retrieve isn't used anywhere besides in its own class (despite public visibility) so you can easily extend the method to take in a CommandSender argument. You're welcome to merge retrieveMessage and formatMessage into one method if you want—both are small and only used together. 
Extending the methods with a CommandSender is a little annoying, granted, but it is semantically correct: properly retrieving a message now depends on the sender.
Keeping modifications minimal is a noble idea, but consistency is important. You've introduced a _global_ replacement tag. If we use the `retrieveMessage` method sometime in the future we'd expect all tags to be replaced, yet this is currently not the case.

Maxetto:
I've tried something but I don't like it really much...
Also, first time integrating Git in Eclipse and writing a test case, is everything correct?

https://github.com/Maxetto/AuthMeReloaded/commit/00de15016a1eeaba4349fc224c275b66f904b28f

ljacqu:
Hi @Maxetto thanks a lot for your efforts! It looks great and the test case you've added is really nice, too! ;)

PR #1512, opened 2018-03-01
Remove VIP status to OP by default

### LINKED ISSUE #1040: Disable VIP-kick
On my server with Authme, when an OP joins (with authme.vip permission), a random person get kicked with the message "A VIP player has joined the server when it was full!".
Is there a way to keep this function while not kicking people?
I'd like the OPs to join even if it's full.

Thanks in advance

Maxetto:
Closes #1040.

I think this is the best solution, better than implementing a case to not kick anyone when an OP player joins.
There's Essentials and other plugins for allowing more people than maximum slots.


sgdc3:
Thank you

PR #1513, opened 2018-03-02
Fix last commit

Maxetto:
I haven't dig enough inside the source code of the new AuthMe and so I always forget about tests.


sgdc3:
Yep, sorry, i didn't check if it was covered by a test xD

Issue #642, opened 2016-04-07
Kick players that tries to register/login too quickly


sgdc3:
Prevents bot attacks, an idea by @games647 ;)

Platinteufel:
Please make this configurable (maybe default=false).

sgdc3:
Sure ;)

HexelDev:
Merged, I'm closing this issue <3

PR #1517, opened 2018-03-04
#642 - Quick Command Protection

HexelDev:
The point of this Protection is simply detect "suspicious" players in order to kick 'em before the anti bot will be triggered.

I've implemented a manager class that:

- Has an `ExpiringSet` used to store the name for the defined time `QUICK_COMMANDS_DENIED_BEFORE_MILLISECONDS`
  - The name is stored only if the player has the permission `PlayerPermission.QUICK_COMMANDS_PROTECTION`
- Checks if the player is in the `Set`
  - Used to know if the player should be kicked

And yeah, that's it...

## Question:

Should the setting `QUICK_COMMANDS_DENIED_BEFORE_MILLISECONDS` be under the `ProtectionSettings` section (the one used by the anti bot stuffs)  or somewhere else?

## TODO:
- [x] Add kick message
- [x] Check and add players to the set on login
- [x] Check and kick players in set on command execution

### DIFF
--- a/src/main/java/fr/xephi/authme/listener/PlayerListener.java
+++ b/src/main/java/fr/xephi/authme/listener/PlayerListener.java
@@ -105,7 +105,7 @@ public class PlayerListener implements Listener {
         final Player player = event.getPlayer();
         if (!quickCommandsProtectionManager.isAllowed(player.getName())) {
             event.setCancelled(true);
-            player.kickPlayer(m.retrieveSingle(MessageKey.QUICK_COMMAND_PROTECTION_KICK));
+            player.kickPlayer(m.retrieveSingle(player, MessageKey.QUICK_COMMAND_PROTECTION_KICK));
             return;
         }
         if (listenerService.shouldCancelEvent(player)) {
--- a/src/main/java/fr/xephi/authme/listener/PlayerListener.java
+++ b/src/main/java/fr/xephi/authme/listener/PlayerListener.java
@@ -1,5 +1,6 @@
 package fr.xephi.authme.listener;
 
+import fr.xephi.authme.data.QuickCommandsProtectionManager;
 import fr.xephi.authme.data.auth.PlayerAuth;
 import fr.xephi.authme.datasource.DataSource;
 import fr.xephi.authme.message.MessageKey;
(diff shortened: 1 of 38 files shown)


ljacqu:
Not much to say just yet, looks good so far ;) 

The idea is that on join you put the user in the set and then if /login or /register happens it will check if the user is still in the set?

Minor comments:
- The javadoc over the new permission node is the same as the one above.
- If the duration config is `0` or negative will this protection be disabled? I think it does and that's nice, since a lot of our settings already behave like that.
- The new setting is called `denyCommandsBeforeSeconds` but it's actually milliseconds. Please update the name (or drop it and specify it in the comment).
- The `permissionManager` field can be final, too.

HexelDev:
@ljacqu 
> The idea is that on join you put the user in the set and then if /login or /register happens it will check if the user is still in the set?

Exactly, but I don't know if we should check also other commands...

@ljacqu 
> If the duration config is 0 or negative will this protection be disabled? I think it does and that's nice, since a lot of our settings already behave like that.

Should we check it when the class is created or just let the set handle it?

ljacqu:
> Should we check it when the class is created or just let the set handle it?

It's probably fine just like that, since it's pretty clear behavior from the set imho :)

HexelDev:
I'm now writing the test class, and I'll generate the mesages etc...

HexelDev:
This should be ready, what do you think about it? 

@ljacqu @sgdc3

sgdc3:
Good for me ;)

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -207,6 +215,11 @@ public void onPlayerJoin(PlayerJoinEvent event) {
         if (!PlayerListener19Spigot.isPlayerSpawnLocationEventCalled()) {
             teleportationService.teleportOnJoin(player);
         }
+
+        if (quickCommandsProtectionManager.shouldSaveLogin(player)) {
+            quickCommandsProtectionManager.setLogin(player.getName());
+        }
```
ljacqu: I would personally merge both methods into something like `processLogin(player)` (or saveLogin or whatever) and the cmd protection manager itself takes care of checking the permission node.

HexelDev: Perfect, editing ;)

PR #1518, opened 2018-03-04
Fixing session expired message spam

sgdc3:


ljacqu:
Looks OK to me. Typically you put a space after `if` because it's a reserved keyword.

Is it really interesting to show any message at all if it's just when the player's IP changes? I expect that typically sessions are configured to live for < 30 minutes and that in this time frame the IP of a user won't change, so the message will probably be shown very rarely now.

That's fine but at that point I would have probably skipped this complex logic and just logged the reason for a failed session to debug or something. :P But I can live with this too.

HexelDev:
As @sgdc3 told me, someone was complaining about the fact that the "session expired" message was shown pretty much on every login (if the session expired). Even if it was an obvious case... (like login after 1 month).

ljacqu:
Thanks for the background, this is fine IMHO and ready to merge if you agree.

HexelDev:
Perfect, merging ;)

PR #1520, opened 2018-03-07
Global tag replacements

Maxetto:
Continuation of #1509, followed @ljacqu remarks and added a Display Name tag too.

I've also added a comment on every translation (translatable from contributors) showing every global replacement that can be done (seems like only FR translation have added it by default).


ljacqu:
Thanks for implementing the remarks from the previous PR. The display name implementation introduces a lot of duplication and duplicated methods. I'll check out the branch so I can make a suggestion. I don't really have the energy right now

ljacqu:
Thanks, merged! I'll have a look at how to fix the duplications later.

Issue #1471, opened 2018-01-11
Migrate to maxmind's Geo-IP 2


games647:
>The free versions of the GeoLite Legacy databases available at https://dev.maxmind.com/geoip/legacy/geolite/ will no longer be updated and free downloads of GeoLite Legacy using the geoipupdate program will no longer be available starting April 1, 2018.

Source: https://support.maxmind.com/geolite-legacy-discontinuation-notice/

The new project is located [here](https://github.com/maxmind/GeoIP2-java), but contains a lot of dependencies. The EssentialsX project noticed that [too](https://github.com/EssentialsX/Essentials/pull/1748).

sgdc3:
We tried to migrate, i remember we had issues with it

sgdc3:
@DNx5 worked on it

sgdc3:
#539

ljacqu:
A check with `InternetProtocolUtils#isLocalAddress` and wrapping the new call in a try-catch should be sufficient to handle the issues we encountered in #539. 

As for the dependency size, either we live with an additional ~1 MB in our JAR file, or we start talking about having individual modules, kind of like Essentials has (you can use Essentials alone or you can add in EssentialsSpawn, etc. for additional features)

sgdc3:
Nah i mamaged to make it download the db correctly for another project, we don't need to include the db

ljacqu:
No one said we need to include the db

sgdc3:
the lib is 1mb? O_o sorry i misread your message xD

k-jiang:
The libs are 1.2mb minimum even with shade-plugin. You can check the [Essentials' PR](https://github.com/EssentialsX/Essentials/pull/1748) for details.
The new Maxmind's Java API requires serval `jackson` libs, which are huge, to work properly. And they use tar.gz to pack their db so you will need `javatar` too.

You can ask my help if you meet any problems. I have used AuthMe for over 6 years and would happy to support it.

sgdc3:
@k-jiang I already implemented the new api in a private project, i think it would be better to provide a separate module, in order reduce the final jar size, it would allow us to remove other huge libraries like the email one. @ljacqu

Maxetto:
I've implemented it too with ease in one of my projects, but the real issue is the Jar size. For me it went from 52 KB to 1.8 MB, with shading.

PR #1529, opened 2018-03-15
Migrate to GEO IP 2 (Related #1471)

games647:
As mentioned in ticket #1471 , MaxMind will stop updating the GEO IP database in April.

Copied from the commit message:

* Update MaxMind database dependency and add javatar to extract the database from the tar archive
(now only a small difference in jar file size)
* Verify downloaded archive using MD5 (There are no other checksums available)
* Migrated to Java NIO instead of old java file I/O (Feedback?)
* Internal Optional usage for nullable values (Feedback?)
* Changed the DatabaseReader at MaxMind to use an interface. This allows us to mock the final class.
* File modification date represents the last time the database was updated at MaxMind.

Minor:
* Change all links of MaxMind to https
* Schedule a Bukkit async task instead of creating a thread manually
* Validate IP input string
* Extract validation into single method
* Close all resources safely using try-resources

Once approved, the commits should be squashed into one and that commit should include the issue number.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -407,7 +407,7 @@
         <dependency>
             <groupId>com.maxmind.db</groupId>
             <artifactId>maxmind-db-gson</artifactId>
-            <version>2.0.2</version>
+            <version>2.0.2-SNAPSHOT</version>
         </dependency>
 
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -403,13 +403,15 @@
             <optional>true</optional>
         </dependency>
 
-        <!-- Maxmind GeoIp API -->
+        <!-- MaxMind GEO IP with our modifications to use GSON in replacement of the big Jackson dependency -->
+        <!-- GSON is already included and therefore it reduces the file size in comparison to the original version -->
         <dependency>
             <groupId>com.maxmind.db</groupId>
             <artifactId>maxmind-db-gson</artifactId>
             <version>2.0.2-SNAPSHOT</version>
         </dependency>
 
+        <!-- Library for tar archives -->
         <dependency>
             <groupId>javatar</groupId>
             <artifactId>javatar</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -267,8 +267,12 @@
                             <shadedPattern>fr.xephi.authme.libs.slf4j.slf4j</shadedPattern>
                         </relocation>
                         <relocation>
-                            <pattern>com.maxmind.geoip</pattern>
-                            <shadedPattern>fr.xephi.authme.libs.maxmind.geoip</shadedPattern>
+                            <pattern>com.maxmind.db</pattern>
+                            <shadedPattern>fr.xephi.authme.libs.maxmind</shadedPattern>
+                        </relocation>
+                        <relocation>
+                            <pattern>com.ice.tar</pattern>
+                            <shadedPattern>fr.xephi.authme.libs.tar</shadedPattern>
                         </relocation>
                         <relocation>
                             <pattern>net.ricecode.similarity</pattern>
@@ -401,11 +405,15 @@
 
         <!-- Maxmind GeoIp API -->
         <dependency>
-            <groupId>com.maxmind.geoip</groupId>
-            <artifactId>geoip-api</artifactId>
-            <version>1.3.1</version>
-            <scope>compile</scope>
-            <optional>true</optional>
+            <groupId>com.maxmind.db</groupId>
+            <artifactId>maxmind-db-gson</artifactId>
+            <version>2.0.2</version>
+        </dependency>
+
+        <dependency>
+            <groupId>javatar</groupId>
+            <artifactId>javatar</artifactId>
+            <version>2.5</version>
         </dependency>
 
         <!-- Java Email Library -->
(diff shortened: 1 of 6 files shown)


src/main/java/fr/xephi/authme/service/GeoIpService.java
```diff
@@ -1,46 +1,80 @@
...
+    private static final String DATABASE_EXT = ".mmdb";
+    private static final String DATABASE_FILE = DATABASE_NAME + DATABASE_EXT;
+
+    private static final String ARCHIVE_FILE = DATABASE_NAME + ".tar.gz";
+
+    private static final String ARCHIVE_URL = "https://geolite.maxmind.com/download/geoip/database/" + ARCHIVE_FILE;
+    private static final String CHECKSUM_URL = ARCHIVE_URL + ".md5";
 
-    private final File dataFile;
+    private static final int UPDATE_INTERVAL = 30;
```
ljacqu: Maybe here add _DAYS to the name to make the unit clear—I had no idea until I saw its usage.

src/main/java/fr/xephi/authme/service/GeoIpService.java
```diff
@@ -49,94 +83,173 @@
...
+
+                // MD5 checksum verification
+                String targetChecksum = Resources.toString(new URL(CHECKSUM_URL), StandardCharsets.UTF_8);
+                if (!verifyChecksum(Hashing.md5(), tempFile, targetChecksum)) {
+                    return;
+                }
+
+                // tar extract database and copy to target destination
+                if (!extractDatabase(tempFile, dataFile)) {
+                    ConsoleLogger.warning("Cannot find database inside downloaded GEO IP file at " + tempFile);
```
ljacqu: Should we `return;` after this or should `downloading` really be set to `false` in this case?

src/main/java/fr/xephi/authme/service/GeoIpService.java
```diff
@@ -49,94 +83,173 @@
...
      * Create a thread which will attempt to download new data from the GeoLite website.
+     */
+    private void startDownloadTask() {
+        downloading = true;
+
+        // use bukkit's cached threads
+        bukkitService.runTaskAsynchronously(() -> {
+            try {
+                // download database to temporarily location
+                Path tempFile = Files.createTempFile(ARCHIVE_FILE, null);
```
ljacqu: Although this is a temp file (and I don't know the best usages) would you oppose to deleting it in a `finally`? I have no idea when an OS would decide it can remove the file again, maybe it's good if we help it along. We have FileUtils#delete for `File` to make it null-safe, maybe we can have a `Path` equivalent.

src/main/java/fr/xephi/authme/service/GeoIpService.java
```diff
@@ -49,94 +83,173 @@
...
     }
 
     /**
      * Create a thread which will attempt to download new data from the GeoLite website.
+     */
+    private void startDownloadTask() {
+        downloading = true;
+
+        // use bukkit's cached threads
+        bukkitService.runTaskAsynchronously(() -> {
```
ljacqu: Would you oppose to logging on INFO that we're downloading a fresh DB at this point?

pom.xml
```diff
@@ -401,11 +405,15 @@
 
         <!-- Maxmind GeoIp API -->
         <dependency>
```
ljacqu: PS: Maybe add a comment for this dependency that it's our adaptation

src/main/java/fr/xephi/authme/service/GeoIpService.java
games647: You're right. I'm going to change that.

src/main/java/fr/xephi/authme/service/GeoIpService.java
games647: Thanks for pointing out. `return` is missing here.

src/main/java/fr/xephi/authme/service/GeoIpService.java
games647: Added

pom.xml
games647: Fixed

src/main/java/fr/xephi/authme/service/GeoIpService.java
games647: Yeah, deleting the file afterwards is better. 

>We have FileUtils#delete for File to make it null-safe, maybe we can have a Path equivalent.

Wouldn't this duplicate existing logic (including tests)?

pom.xml
ljacqu: Awesome!

src/main/java/fr/xephi/authme/service/GeoIpService.java
ljacqu: > Wouldn't this duplicate existing logic

That's a good argument and your solution is fine. You could have also argued that we should use nio types like `Path` more and then we'd have this util already handy. Doesn't really matter to me personally :wink:

ljacqu:
@sgdc3 This looks fantastic to me – do you have a comment to add? If not please merge

sgdc3:
Looks perfect!

Issue #1531, opened 2018-03-16
NoSuchMethodError: org.bukkit.Bukkit.spigot()


Eufranio:
When starting my 1.7.10 server with AuthMe, I get this error: 
```
[15:33:34] [Server thread/ERROR]: Error occurred while enabling AuthMe v5.5.0-SNAPSHOT-b2027 (Is it up to date?)
java.lang.NoSuchMethodError: org.bukkit.Bukkit.spigot()Lorg/bukkit/Server$Spigot;
at fr.xephi.authme.settings.SettingsWarner.logWarningsForMisconfigurations(SettingsWarner.java:57) ~[?:?]
at fr.xephi.authme.AuthMe.onEnable(AuthMe.java:150) ~[?:?]
at org.bukkit.plugin.java.JavaPlugin.setEnabled(JavaPlugin.java:316) ~[JavaPlugin.class:1.7.10-1614.58]
at org.bukkit.plugin.java.JavaPluginLoader.enablePlugin(JavaPluginLoader.java:368) [JavaPluginLoader.class:1.7.10-1614.58]
at org.bukkit.plugin.SimplePluginManager.enablePlugin(SimplePluginManager.java:410) [SimplePluginManager.class:1.7.10-1614.58]
at org.bukkit.craftbukkit.v1_7_R4.CraftServer.loadPlugin(CraftServer.java:438) [CraftServer.class:1.7.10-1614.58]
at org.bukkit.craftbukkit.v1_7_R4.CraftServer.enablePlugins(CraftServer.java:372) [CraftServer.class:1.7.10-1614.58]
at net.minecraft.server.MinecraftServer.func_71243_i(MinecraftServer.java:519) [MinecraftServer.class:?]
at net.minecraft.server.MinecraftServer.func_71222_d(MinecraftServer.java:493) [MinecraftServer.class:?]
at net.minecraft.server.MinecraftServer.func_71247_a(MinecraftServer.java:453) [MinecraftServer.class:?]
at net.minecraft.server.dedicated.DedicatedServer.func_71197_b(DedicatedServer.java:337) [lt.class:?]
at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:643) [MinecraftServer.class:?]
at java.lang.Thread.run(Thread.java:748) [?:1.8.0_151]
```

sgdc3:
@Eufranio Should be fixed, let me know

PR #1534, opened 2018-03-19
#1531 Move spigot detection to BukkitService

ljacqu:
Moved the logic for Spigot to BukkitService so the method can be mocked


ljacqu:
hahaha!

Issue #1141, opened 2017-03-24
2FA: Make two-factor optional


ljacqu:
Follow-up of #87, we currently have a `TWO_FACTOR` implementation but it forces everyone to use it. In this issue, we want to make it optional:

- if enabled, the player should be able to _decide_ whether he wants to use a password, or 2FA -> requires a new command
- internally, this means the player can decide between two hash algorithms. For this reason, the easiest way to keep track of this would be to add a new hash_algorithm column to the database. 

### Open question
With this new setup it should not be possible for plugin admins to configure `TWO_FACTOR` as hash algorithm. What happens if someone currently has this? This requires some smart migration...

Jamaic230:
I think 2nd factor authentication should be "separate" from login.

Internally, each user would have a column for OTP secret. If _null_, he's authenticated by password (token parameter is ignored), otherwise he's authenticated with both, password and token.

Scenario - User logs in without 2FA:
> 1) User joins
> 2) User types: /login <validPassword>
> 3) User is logged in

Scenario - User logs in with 2FA:
> 1) User joins
> 2) User types: /login <validPassword> <validToken>
> 3) User is logged in

Scenario - User turns 2FA on:
> 1) User is logged in
> 2) User types: /2fasetup
> 3) Server responses with OTP secret (qrcode, link to website, etc.)
> 4) User types: /2fasetup <validToken>
> 5) User now have 2FA active

Scenario - User turns 2FA off:
> 1) User is logged in
> 2) User types: /2faremove <validToken>
> 3) User now have 2FA inactive

Changing/recovering passwords would not be affected by 2FA.

ljacqu:
@Jamaic230 Thanks for your input, this is all fine by me. With an admin command to turn 2FA off for a user we cover everything.

ljacqu:
I move this to 5.5 with a somewhat "heavy heart" since I know that people do want this, but we need to eventually come to a closure with 5.4 and this requires major changes. We'll make this a priority in 5.5.

ljacqu:
Analysis based on @Jamaic230's comment: 2FA should be an additional layer of authentication to the regular password. This means:

- new field on PlayerAuth and in DB
  * If column is NULL player doesn't have 2fa, otherwise yes
  * Could have a follow-up issue to make the column optional maybe. I think once we figure out the specifics regarding configurations / permissions we'll see whether this makes sense or not.
- new command /2fa to enable and disable 2 factor authentication
[...]

PR #1537, opened 2018-03-21
#1141 optional additional 2fa auth

ljacqu:
Not finished yet, but this is the implementation for #1141. Tests are still missing, I want to add some Javadoc, and a whole buttload of testing is still pending :smile: But I'd still love some comments so far.


src/main/java/fr/xephi/authme/service/BukkitService.java
```diff
@@ -393,4 +393,11 @@ public BanEntry banIp(String ip, String reason, Date expires, String source) {
             return Optional.empty();
         }
     }
+
+    /**
+     * @return the IP string that this server is bound to, otherwise empty string
+     */
+    public String getIp() {
+        return Bukkit.getServer().getIp();
```
games647: Maybe this should be configurable in AuthMe to give it a more unique name that isn't a domain or IP address?

pom.xml
```diff
@@ -469,6 +473,13 @@
             <optional>true</optional>
         </dependency>
 
+        <!-- TOTP client -->
+        <dependency>
+            <groupId>com.warrenstrange</groupId>
+            <artifactId>googleauth</artifactId>
+            <version>1.1.2</version>
```
games647: Maybe exclude `org.apache.httpcomponents:httpclient` here? This shouldn't break anything, because we don't use the web API adapter.

src/main/java/fr/xephi/authme/security/totp/TotpAuthenticator.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+public class TotpAuthenticator {
+
+    private final IGoogleAuthenticator authenticator;
+    private final BukkitService bukkitService;
+
+
+    @Inject
+    TotpAuthenticator(BukkitService bukkitService) {
+        this(new GoogleAuthenticator(), bukkitService);
```
games647: This creates by default Base32 keys with SHA1. Since SHA1 is broken we should use something stronger or do you think it's not relevant in this case?

pom.xml
games647: `commons-codec:commons-codec` (for Base32 and Base64 encoding/decoding) is also present in Minecraft since 1.7.10, but in a different version (1.9 vs 1.10) and therefore could maybe break something.

pom.xml
```diff
@@ -469,6 +473,13 @@
...
+        </dependency>
```
games647: An `<optional>true</optional>` would be great here that plugins which depend on AuthMe don't resolve this dependency.

src/main/java/fr/xephi/authme/service/BukkitService.java
ljacqu: Absolutely, that's a great point! Maybe in a further iteration so I can hopefully finish this PR before I leave for vacation, which is very soon.

pom.xml
ljacqu: Good point.

src/main/java/fr/xephi/authme/security/totp/TotpAuthenticator.java
ljacqu: I can try to change it and I'll check if Google Authenticator still works with it. As long as it does it's definitely a great improvement. Thanks for noticing!

pom.xml
ljacqu: Of course! I just lazily added it to the POM until now to focus on the code impl (there are also no shading relocation rules yet). Keen eye ;)

pom.xml
ljacqu: As an update, looks like httpclient is required when generating the QR code with Google :( https://postimg.org/image/68bv03bz1/

ljacqu: What's your suggestion for the commons-codec thing? Should we really exclude it? This would mean that for Sponge etc. we'd also rely on it being there for everything to work.

games647: >As an update, looks like httpclient is required when generating the QR code with Google :( https://postimg.org/image/68bv03bz1/

Okay, my fault sorry. I always generated the URLs on my own. It makes more sense to use the provided method than creating something custom. So looks like we have to shade `httpclient` and `httpcore` then. 

>This would mean that for Sponge etc. we'd also rely on it being there for everything to work.

Good point. 

Base32 and Base64 encoding is also available in guava which we already shade into the final jar, but it would be ridiculous to fork every project that includes dependencies we do not have.

So yeah, then we have to shade it too.

sgdc3:
@games647 is it ok now?

sgdc3:
@HexelDev could you take a look?

pom.xml
```diff
@@ -299,6 +295,10 @@
                             <pattern>org.bstats</pattern>
                             <shadedPattern>fr.xephi.authme.libs.org.bstats</shadedPattern>
                         </relocation>
+                        <relocation>
```
sgdc3: duplicate entry

MineSAT:
2FA auth with google authenticator? (Has API)

ljacqu:
Unfortunately had to revert to using SHA1 again (hi @games647) :( Using SHA512 didn't work with the Google authenticator (the codes it showed didn't correspond with what the server is expecting).

Still open points before merging:
- Migration of existing messages at `misc.two_factor_create` to `two_factor.code_created`
- Unit tests
- Merge TotpService into TotpAuthenticator

I'll create a separate issue for games647 comment:
- Create a configuration to override the server IP/name shown in the authenticator

PR #1547, opened 2018-03-29
[Discussion] Skip IP check for API registration

games647:
Open for discussion:

**Background:**
The user `OrangeFox` suggested in Discord channel `#support` that we should save accounts by their UUID and not by their username, because usernames can now be changed. This way players don't have to register again. Although other plugins can detect name changes, they would **fail to re-register an account at AuthMe, because of the IP limit.**

**Open points about this solution**

* Should a plugin even be able to bypass this limit? Maybe we should introduce an configuration option to let the user decide
* What about disabled registration? Should this option be bypassed too?

**Other solutions:**

* Other plugins should use the offline registration method `registerPlayer(String username, String password)` that doesn't perform validation checks before registration
https://github.com/AuthMe/AuthMeReloaded/blob/ee41d92216e31ddb14bfaba013546c20f874fdc5/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java#L231
     * Registration IP will be missing
https://github.com/AuthMe/AuthMeReloaded/blob/f752b9d7a71b6bc0ee572f59ff54f8165abc5f15/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java#L240-L241
     * "Please add your email", commands on register, BungeeCord server forwarding have to be implemented manually
     * -> Conclusion many AuthMe implementation dependent things
https://github.com/AuthMe/AuthMeReloaded/blob/f752b9d7a71b6bc0ee572f59ff54f8165abc5f15/src/main/java/fr/xephi/authme/process/register/ProcessSyncPasswordRegister.java#L58-L81
* Introduce a way to migrate the account
    * API method to get the PlayerAuth instance, `setName` and then a save method
         * Exposes internal data structures?
    * Migrate username API method to keep all existing data **(My favorite way)**
    * Add a delete method to the API in order to delete the old account and then plugins could register a new account
        * Old data will still be lost (Registration date, IP, ...)


sgdc3:
@ljacqu @Xephi i really like the idea of a migrate api, it would be the cleanest solution in my opinion, what do you think guys?

ljacqu:
This kind of breaks existing behavior but the change in this PR is OK IMHO.

What I could imagine for our next API version is to draw up interfaces of many things we want to expose. As long as we only make interfaces public and make educated choices about our API I think we could then have something very flexible and longstanding. 

I'm a big fan of passing in options as varargs (like java nio Files#write has) as to have only one method for, say, `saveAccount` while still allowing some control over it. This would solve questions like consider if registration is disabled? Check player's country? etc.

sgdc3:
@ljacqu so, is it ok for you if i merge this? ;)

ljacqu:
Absolutely, went ahead and merged :)

PR #1549, opened 2018-04-02
Added EmailChangedEvent

toxuin:
I've needed such events for integration with 3rd party services that use DB layout not really compatible with AuthMe.
My use case is that for every email change, this 3rd party service needs to send off it's own activation email which it stores in own DB.
If AuthMe would have such event, I could write a small tiny plugin that would talk to such 3rd party by, for example, REST API. This way, I could keep emails in sync.

Tests included.

Please let me know what you think.

### DIFF
--- a/src/main/java/fr/xephi/authme/events/EmailChangedEvent.java
+++ b/src/main/java/fr/xephi/authme/events/EmailChangedEvent.java
@@ -2,10 +2,13 @@ package fr.xephi.authme.events;
 
 import org.bukkit.entity.Player;
 import org.bukkit.event.Cancellable;
+import org.bukkit.event.Event;
 import org.bukkit.event.HandlerList;
 
+import javax.annotation.Nullable;
+
 /**
- * This event is called when a player changes his email address.
+ * This event is called when a player adds or changes his email address.
  */
 public class EmailChangedEvent extends CustomEvent implements Cancellable {
     private static final HandlerList handlers = new HandlerList();
@@ -14,7 +17,16 @@ public class EmailChangedEvent extends CustomEvent implements Cancellable {
     private final String newEmail;
     private boolean isCancelled;
 
-    public EmailChangedEvent(Player player, String oldEmail, String newEmail, boolean isAsync) {
+    /**
+     * Constructor
+     *
+     * @param player The player that changed email
+     * @param oldEmail Old email player had on file. Can be null when user adds an email
+     * @param newEmail New email that player tries to set. In case of adding email, this will contain
+     *                the email is trying to set.
+     * @param isAsync should this event be called asynchronously?
+     */
+    public EmailChangedEvent(Player player, @Nullable String oldEmail, String newEmail, boolean isAsync) {
         super(isAsync);
         this.player = player;
         this.oldEmail = oldEmail;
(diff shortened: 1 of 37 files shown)


src/main/java/fr/xephi/authme/events/EmailChangedEvent.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public boolean isCancelled() {
+        return isCancelled;
+    }
+    
+    public Player getPlayer() {
+        return player;
+    }
+
+    public String getOldEmail() {
```
games647: Please document nullable values (including when/why) like here if the user adds a new email address.

src/main/java/fr/xephi/authme/process/email/AsyncAddEmail.java
```diff
@@ -57,14 +62,19 @@ public void addEmail(Player player, String email) {
             } else if (!validationService.isEmailFreeForRegistration(email, player)) {
                 service.send(player, MessageKey.EMAIL_ALREADY_USED_ERROR);
             } else {
+                EmailChangedEvent event = bukkitService.createAndCallEvent(isAsync
+                    -> new EmailChangedEvent(player, null, email, isAsync));
+                if (event.isCancelled()) {
+                    sendFailedMessage(player);
```
games647: In my opinion, this makes it hard to track down errors. `dataSource.updateEmail` will print an exception if an error occurrs, but if the event is cancelled you only get a very abstract error message.

My suggestion is to use a different error message for the console where you can see it's because the event was cancelled.

src/main/java/fr/xephi/authme/process/email/AsyncChangeEmail.java
```diff
@@ -57,14 +62,21 @@ public void changeEmail(Player player, String oldEmail, String newEmail) {
...
             outputUnloggedMessage(player);
         }
     }
 
-    private void saveNewEmail(PlayerAuth auth, Player player, String newEmail) {
+    private void saveNewEmail(PlayerAuth auth, Player player, String oldEmail, String newEmail) {
+        EmailChangedEvent event = bukkitService.createAndCallEvent(isAsync
+            -> new EmailChangedEvent(player, oldEmail, newEmail, isAsync));
+        if (event.isCancelled()) {
+            service.send(player, MessageKey.ERROR);
```
games647: Same here.

ljacqu:
This is quality work!

As @games647 already noted please distinguish both cases when logging. If changing an email fails because the event is canceled, the log level shouldn't be WARN—we reserve that for "unexpected" situations that should be investigated.

toxuin:
Hey, thanks for the kind words.

I've added some documentation for the new event. Now, depending on action performed (change\add), it would fire different messages off to user and console.

Well and since I'm at it, fixed up the TODOs in messages_ru.yml. Ура!

src/main/java/fr/xephi/authme/process/email/AsyncAddEmail.java
```diff
@@ -57,6 +62,13 @@ public void addEmail(Player player, String email) {
             } else if (!validationService.isEmailFreeForRegistration(email, player)) {
                 service.send(player, MessageKey.EMAIL_ALREADY_USED_ERROR);
             } else {
+                EmailChangedEvent event = bukkitService.createAndCallEvent(isAsync
+                    -> new EmailChangedEvent(player, null, email, isAsync));
+                if (event.isCancelled()) {
+                    ConsoleLogger.warning("Could not add email to player '" + player + "' – event was cancelled");
```
games647: Like @ljacqu said, warning is reserved for unexpected situation that should be investigated. Please change it to something else like info.

src/main/java/fr/xephi/authme/process/email/AsyncChangeEmail.java
```diff
@@ -57,14 +63,22 @@ public void changeEmail(Player player, String oldEmail, String newEmail) {
...
             outputUnloggedMessage(player);
         }
     }
 
-    private void saveNewEmail(PlayerAuth auth, Player player, String newEmail) {
+    private void saveNewEmail(PlayerAuth auth, Player player, String oldEmail, String newEmail) {
+        EmailChangedEvent event = bukkitService.createAndCallEvent(isAsync
+            -> new EmailChangedEvent(player, oldEmail, newEmail, isAsync));
+        if (event.isCancelled()) {
+            ConsoleLogger.warning("Could not change email for player '" + player + "' – event was cancelled");
```
games647: Same here

sgdc3:
@toxuin Thank you very much for your contribution ;)

Issue #1561, opened 2018-04-20
Timing Side-Channel in RoyalAuth.comparePassword(String, String)


nolleryc:
We have found a **timing side-channel** in class _fr.xephi.authme.security.crypts.RoyalAuth_, in the inherited method "public boolean comparePassword(String password, HashedPassword hashedPassword, String name)". This is due to the use of String.equals for comparison which returns as soon as a character does not match. This represents a timing side channel, which could be used by a potential attacker to obtain knowledge about the hidden secret password.
Do you agree with our findings?

A similar issue is present in several "comparePassword" methods, e.g., in the classes **fr.xephi.authme.security.crypts.Sha256** and **fr.xephi.authme.security.crypts.Pbkdf2**.
We found these classes in the latest version of your GitHub repo.

This vulnerability can be easily avoided by using a safe String comparison like in method fr.xephi.authme.security.crypts.BCryptService.checkpw(String, String).
Do you agree with our patch proposal?

Please feel free to contact us for further clarification! You can reach us by the following email address:
yannic.noller@informatik.hu-berlin.de

Best regards,
Yannic Noller

sgdc3:
Hello, we'll try to address that asap, thank you for the report.

@ljacqu hello, how are you? If you have time could we talk about that on discord?

sgdc3:
@nolleryc Fixed in the latest commits, thank you again for the report, have a nice day ;)

nolleryc:
Great! Thanks for fixing this.

Regarding your questions:

1. The standard string comparison itself is vulnerable and should be not be used in such a context. While this vulnerability might not be exploitable by itself, it can be combined with other techniques (e.g. attacks against hash functions) to reveal secrets; it is the safe way to fix the string comparisons.

1. I agree that in your case this issue might not be critical. But as you never know what a potential attacker might be capable of, i.e. what kind of insights he/she has of your environment, it is the safe way to fix such issues. Since your code is open-source, the attacker will be aware of this vulnerability and will try to use it.

Thanks
Yannic

PR #1563, opened 2018-04-22
#1561 Fix timing attacks by comparing hashes in constant time

ljacqu:
Fix for #1561, basically just replaces `String#equals` calls with a wrapped call to `MessageDigest#isEqual`. I'd be open to a more specific name than `isEqual` to convey the method's very specific intention but everything I came up with sounded... clunky.

cc: @nolleryc


ljacqu:
Thank you! It is now fixed

PR #1570, opened 2018-05-13
Update messages_pt.yml

rafael59r2:


ljacqu:
That was fast! Thank you :)

Issue #1524, opened 2018-03-09
2FA: Migrate the TWO_FACTOR hash algorithm


ljacqu:
Unfortunately we cannot perform our usual way of migrations with the TWO_FACTOR hash algorithm since it means a TOTP key is stored in the database.

I'd suggest to migrate the current TWO_FACTOR password (on success) to the new totpKey column and to set some predefined password (ew, I know) to the user. Unless we can _force_ the user to set a new password right after a successful TWO_FACTOR code was input.

Ensure that TWO_FACTOR can no longer be set as `hash` config in the config.yml file.

PR #1576, opened 2018-05-20
#1524 Drop old TWO_FACTOR hash algorithm, force new password to be input

ljacqu:
Deprecates TWO_FACTOR from being configured as the hash algorithm and forces players with a TWO_FACTOR password to set a new password with /register when they log in. I don't know how many servers actually use this so I went with kicking a user after successfully setting a password. 

Open question:
- Migrate the two factor secret to the new totp column?


games647:
There could be some server that use this method for 2FA in addition to Mojang authentication, but maybe in this case it might be better to choose a dedicated 2FA plugin or what do you think?

ljacqu:
Hmm that's a good point. We could have a "mode" in the future that would specify when/how we require a password or a 2FA token, or at least configurable in the config.yml. But for now I'd like to remove the TWO_FACTOR hash method because it's not a hash method and so has various issues (can't migrate away from it as with the other hashes, /changepassword doesn't work, ...)

sgdc3:
@ljacqu any news?

ljacqu:
Alright, picking this one up again and making it happen in the course of the week ;)

sgdc3:
@ljacqu just as a reminder ;)

PR #1592, opened 2018-06-24
Add IPv6 support for isLocal checks

games647:
* Replace magic values like 127.0.0.1 and use our utility
* Support for IPv6 local adresses in IPv6 only or dual stack environments
    * Loopback [::1]
    * Site-Local fc00::/7
    * Link-local fe80::/10

For site-local I had to introduce a workaround, because Java only checks for deprecated site-local addresses. 

Furthermore I added some documentation that should clear some things up. Reference comments in cd4693eedf60878b184144a8dbdf27d2443fe726.


sgdc3:
Looks very good but tests are still failing.

games647:
I believe this in an incorrect test, because 127.0.0.4 is a loopback address and therefore returns false without calling the getOnlinePlayers().

sgdc3:
Yeah i think so too, but i'm currently without pc and i can't fix it, could you take a look? :)

sgdc3:
@ljacqu could you take a look why test are failing?

ljacqu:
@sgdc3 I'm on it!

PR #1593, opened 2018-06-25
Update translations.md

rafael59r2:


ljacqu:
We auto-generate the file (as the comment on top says and the footer on the page) so what we can do is run that task since this page seems to be a little behind.

PR #1614, opened 2018-08-16
Fix messaging [WIP]

sgdc3:
Messages are not properly forwarded to the bungeecord network.

TODO:
- [x] make login and logout broadcast messages (involves updating AuthMeBungee probably)
- [x] move the (database instanceof CacheDataSource) check login into a proper service (maybe? feedback pls)

### DIFF
--- a/src/main/java/fr/xephi/authme/process/changepassword/AsyncChangePassword.java
+++ b/src/main/java/fr/xephi/authme/process/changepassword/AsyncChangePassword.java
@@ -3,7 +3,6 @@ package fr.xephi.authme.process.changepassword;
 import fr.xephi.authme.ConsoleLogger;
 import fr.xephi.authme.data.auth.PlayerAuth;
 import fr.xephi.authme.data.auth.PlayerCache;
-import fr.xephi.authme.datasource.CacheDataSource;
 import fr.xephi.authme.datasource.DataSource;
 import fr.xephi.authme.message.MessageKey;
 import fr.xephi.authme.process.AsynchronousProcess;
--- a/src/main/java/fr/xephi/authme/process/changepassword/AsyncChangePassword.java
+++ b/src/main/java/fr/xephi/authme/process/changepassword/AsyncChangePassword.java
@@ -55,9 +55,7 @@ public class AsyncChangePassword implements AsynchronousProcess {
                 commonService.send(player, MessageKey.ERROR);
                 return;
             }
-            if (dataSource instanceof CacheDataSource) {
-                bungeeSender.sendAuthMeBungeecordMessage(MessageType.REFRESH_PASSWORD, name);
-            }
+            bungeeSender.sendAuthMeBungeecordMessage(MessageType.REFRESH_PASSWORD, name);
 
             playerCache.updatePlayer(auth);
             commonService.send(player, MessageKey.PASSWORD_CHANGED_SUCCESS);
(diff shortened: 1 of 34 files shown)


src/main/java/fr/xephi/authme/service/bungeecord/BungeeSender.java
```diff
@@ -75,13 +75,17 @@ public void connectPlayerOnLogin(Player player) {
...
-            if(!plugin.isEnabled()) {
+            if (!plugin.isEnabled()) {
                 ConsoleLogger.debug("Tried to send a " + type + " bungeecord message but the plugin was disabled!");
                 return;
             }
-            sendBungeecordMessage("AuthMe", type, playerName.toLowerCase());
+            if (type.isBroadcast()) {
+                sendBungeecordMessage("Forward", "ALL", "AuthMe", type.getId(), playerName.toLowerCase());
+            } else {
+                sendBungeecordMessage("AuthMe", type.getId(), playerName.toLowerCase());
```
games647: This won't work in [1.13. Channels have to be in the following format:](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/browse/src/main/java/org/bukkit/NamespacedKey.java#32)
Namespace, Key := [a-z0-9._-]
Channel := Namespace:Key

and for 1.7.X support [smaller than 16 characters in total](https://github.com/bergerkiller/CraftSource/blob/8e1126adbbde1096b8351650bdd2c1ca0f6a9e46/org.bukkit/plugin/messaging/Messenger.java#L21).

src/main/java/fr/xephi/authme/process/changepassword/AsyncChangePassword.java
```diff
@@ -84,7 +87,9 @@ public void changePasswordAsAdmin(CommandSender sender, final String playerName,
 
         HashedPassword hashedPassword = passwordSecurity.computeHash(newPassword, lowerCaseName);
         if (dataSource.updatePassword(lowerCaseName, hashedPassword)) {
-            bungeeSender.sendAuthMeBungeecordMessage(MessageType.REFRESH_PASSWORD, lowerCaseName);
+            if (dataSource instanceof CacheDataSource) {
+                bungeeSender.sendAuthMeBungeecordMessage(MessageType.REFRESH_PASSWORD, lowerCaseName);
+            }
```
ljacqu: 1) Maybe it‘s more elegant to have the bungee sender take care of checking that if statement. The method could be named accordingly. 

2) Can we introduce a method on DataSource for this? Something like `boolean isCached()` or `boolean hasCache()`

src/main/java/fr/xephi/authme/service/bungeecord/MessageType.java
```diff
@@ -1,19 +1,51 @@
...
+
+    /**
+     * Returns the MessageType with the given ID.
+     *
+     * @param id the message type id.
+     * @return the MessageType with the given id, MessageType.UNKNOWN if invalid.
+     */
+    public static MessageType fromId(String id) {
+        for(MessageType current : values()) {
+            if(current.getId().equals(id)) {
```
ljacqu: Minor: Please add space before reserved keywords like `if` and `for`

src/main/java/fr/xephi/authme/service/bungeecord/MessageType.java
```diff
@@ -1,19 +1,51 @@
...
+                return current;
+            }
+        }
+        return UNKNOWN;
```
ljacqu: Discussion: does it make sense to return an enum entry here that I guess only exists for this method? Other alternatives could be to return `null` here, or an Optional

sgdc3: not a big fun of nulls, and an optional is a little overkill in this case, idk

src/main/java/fr/xephi/authme/service/bungeecord/MessageType.java
ljacqu: Good point; I think there‘s a Checkstyle rule that can be added. I can check once I‘m back home :) (still on vacation)

src/main/java/fr/xephi/authme/service/bungeecord/MessageType.java
ljacqu: I saw you changed it – thank you – the reason it irked me is that it‘s an additional entry in the enum but it‘s not really a message type; it‘s a marker for an invalid state, so to speak.

sgdc3:
@ljacqu, any idea why the jdk8 circle build is failing? :/

ljacqu:
The internet here in Kyrgyzstan is so bad I can‘t even load the CircleCI log 😂

PR #1620, opened 2018-08-25
Add PostgreSQL support

Alan-Gomes:


sgdc3:
https://codeclimate.com/github/AuthMe/AuthMeReloaded/pull/1620

Alan-Gomes:
@sgdc3 
1- I've tested locally with the latest version of postgres (10.5) and it works like a charm

2- These warnings came from methods that I've copied from the MySQL class without any changes, that's weird

sgdc3:
The parameters are simply ignored, please adapt them accordling to https://jdbc.postgresql.org/documentation/head/connect.html

ljacqu:
I‘m not sure if it‘s possible with postgres but it would be awesome if we could have an integration test for the postgres variant as we do for the others.

Alan-Gomes:
@sgdc3 I fixed the jdbc parameters according to the docs, some of them are'nt available on postgres :/

Alan-Gomes:
@ljacqu I now created the tests according to the pattern of the project, I think it's right

Alan-Gomes:
code climate is showing issues in methods that I'm not the author, I don't pretend to fix them to not create regression problems

ljacqu:
To me everything looks great! Thanks for adding the tests. I think abstracting more of the common things between Postgres and MySQL isn't worth it since we already have quite a hierarchy in the data sources, plus we can't really do it in a natural way as soon as there is even a _minimal_ difference in syntax or parameters to set. 
Totally agree on not fixing issues you haven't introduced per se; in fact later on I wouldn't be against just suppressing those errors for the methods that are too long.

@sgdc3 The only thing I worry about is the addition of the postgres dependency to the pom? What's our strategy there?

sgdc3:
@ljacqu sorry, what do you mean?

ljacqu:
@sgdc3 Now postgres is a new dependency in the pom and I guess it will be shaded in, always. Is that an issue? I'm not on top of the POM stuff :sweat_smile:

sgdc3:
We need to relocate it into our package, our jar is becoming too large, we should start working on a runtime dependency resolver.

ljacqu:
OK, agree that we're starting to have an issue, shall we collect some idea and discuss sometime? I want to explore options before we start doing anything "weird" :D

sgdc3:
@ljacqu sure! ;)

pom.xml
```diff
@@ -808,6 +808,12 @@
             <optional>true</optional>
         </dependency>
 
+        <dependency>
```
sgdc3: Please relocate the driver and its dependencies into our package in the maven shade plugin configuration.

Alan-Gomes: Done :)

sgdc3:
@Alan-Gomes Thank you for your contribution! :)

Issue #1623, opened 2018-08-28
sendPlayerTo suggestion


MineyShadow:
Maybe there could be a permission that prevents a player from being sent to the server defined in the configuration? This is really useful for those who work with a seperate server for authentication, which get automatically sent to the main server after login.

sgdc3:
@MineyShadow could you please provide an example scenario?

MineyShadow:
Two servers, one being the lobby and the other being the authentication server where players join in, After they login, they are automatically sent to the lobby if the configuration setting is used, however perhaps there are "staff" that want to work on something within the authentication server [Such as building] but get sent to the lobby after login, the only workaround i know of is disabling the feature in config, /authme reload, doing whatever the staff does then turning the feature back on. Paired with the plugin fastlogin also means premium users get automatically registered/logged in therefore sent to the lobby server. The "permission" i mentioned above can be applied on a player which will prevent authme from sending the player to the lobby.

ljacqu:
Hi @MineyShadow , this has been implemented by @Alan-Gomes and is available as of build nr. 2114: https://ci.codemc.org/job/AuthMe/job/AuthMeReloaded/ . Can you give us a quick feedback whether this issue is OK to close?

MineyShadow:
Testing on spigot 1.12.2 with the aforementioned authme reloaded build along with luckperms 4.2.82 shows that the new feature is working without any error, Thanks for the awesome feature! you may close this issue now.

ljacqu:
Thanks for the feedback!

PR #1626, opened 2018-09-01
Create permission to bypass bungeecord server teleportation

Alan-Gomes:
Based on issue #1623 I created a permission to allow administrators to bypass the bungeecord server teleportation after login.

The permission is **authme.bypassbungeesend**


Alan-Gomes:
I was not sure which is the best name, I thought of "authme.bungeesend" or "authme.bungeeteleportation" so I chose the shortest one

sgdc3:
Thank you

Issue #1627, opened 2018-09-02
Replace BCryptService with Maven dependency


ljacqu:
Replace the BCryptService we've copied with a Maven dependency to https://github.com/patrickfav/bcrypt

The code seems more matured and gets rid of issues we've had to manually fix (such as support for 2y hashes).

PR #1629, opened 2018-09-02
#1627 Replace BCryptService with Maven dependency

ljacqu:
- Remove BCryptService in favor of a better BCrypt implementation (Maven dependency)
- Introduce BCryptHasher wrapping the dependency with more suitable methods
- Fix inaccurate details about salt length in docu annotation: for BCrypt it's always 22 chars
- Change phpBB hash to produce 2y hashes instead of 2a


src/main/java/fr/xephi/authme/security/crypts/BCryptHasher.java
```diff
@@ -0,0 +1,75 @@
...
+     * @param costFactor the log2 cost factor to use
+     */
+    public BCryptHasher(BCrypt.Version version, int costFactor) {
+        this.hasher = BCrypt.with(version);
+        this.costFactor = costFactor;
+    }
+
+    public HashedPassword hash(String password) {
+        byte[] hash = hasher.hash(costFactor, password.getBytes());
+        return new HashedPassword(new String(hash));
```
ljacqu: woops, this relies on the default encoding

PR #1631, opened 2018-09-09
Update configme

ljacqu:
Updates ConfigMe to 1.0.1. https://github.com/AuthMe/ConfigMe/blob/master/CHANGELOG.md

Haven't checked the plugin in Minecraft but tests run and I've run the tool tasks that have been changed.

Edit: one test fails now; I'll fix it -.-


sgdc3:
:+1:

PR #1633, opened 2018-09-09
Do not create a welcome.txt file when boolean is false

montlikadani:
It would be nice if you would accept it as if the boolean is false and creates a file, but this is not good and unnecessary. Therefore, this should not create a file.


ljacqu:
Hi, great idea, but there appears to be acompile error.


> [ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/circleci/authmereloaded-jdk8/src/main/java/fr/xephi/authme/settings/WelcomeMessageConfiguration.java:[116,77] incompatible types: missing return value
[INFO] 1 error

sgdc3:
Tests are failing due to missing service mock.

PR #1634, opened 2018-09-09
Remove 1.7 support

sgdc3:
Remove 1.7 support, will reduce jar size.

@ljacqu could you take a look at this and maybe update the limbo cache system to use UUIDs instead of strings?


sgdc3:
@ljacqu any idea why tests are failing?

ljacqu:
Changes look fine, I just wonder if dropping 1.7 is really something we want to do? It doesn't drop _that much_ legacy code and on Discord it seemed like some people had quite some emotions about it. If reducing the JAR size is our ultimate goal could we try to solve that some other way and leave this on the side for now?

>  update the limbo cache system to use UUIDs instead of strings?

If we want to keep the data we'd have to migrate to another form of storage probably. Would definitely be worth its own issue :P

> why tests are failing

should be fixed now hopefully

Hex3l:
> If reducing the JAR size is our ultimate goal could we try to solve that some other way and leave this on the side for now?

We could achieve that by developing a libraries loader that will be downloaded at runtime ^^

pom.xml
```diff
@@ -73,7 +73,7 @@
         <bukkitplugin.authors>Xephi, sgdc3, DNx5, timvisee, games647, ljacqu, Gnat008</bukkitplugin.authors>
 
         <!-- Change Bukkit Version HERE! -->
-        <bukkit.version>1.13-R0.1-SNAPSHOT</bukkit.version>
+        <bukkit.version>1.13.1-R0.1-SNAPSHOT</bukkit.version>
```
lifehome:
```suggestion
        <bukkit.version>1.13.2-R0.1-SNAPSHOT</bukkit.version>
```
Perhaps catch up the version before merging?

sgdc3:
Outdated

PR #1676, opened 2018-10-31
Remove second argument from /email add

ljacqu:


ljacqu:
Wrong base :) Sorry.

7wOv6ySCjo:
> Wrong base :) Sorry.

why not allow other commands like "/email add" be optional the second argument?
Some servers prefer simplicity in the commands.

Issue #1703, opened 2018-12-05
2FA: "/2fa code <code>" should not be logged in the console [Suggestion]


austinhuang0131:
#### Before reporting an issue make sure you are running the latest build of the plugin and checked for duplicate issues!
yes

### What behaviour is observed:
Console logs `/2fa code <code>` message as usual.

### What behaviour is expected:
Console logs something like "2fa success" instead of exposing the code, just like how `/login` is treated in the console. I'm not saying I don't trust my OPs but I just think it's necessary.

### Steps/models to reproduce:
You login using 2fa.

### Plugin list:
n/a

### Environment description
Spigot 1.13.2 @ Linux

### AuthMe build number:
v5.5.1-SNAPSHOT (build: 2181)

### Error Log:
n/a

### Configuration:
n/a

lifehome:
This indeed should not appear anywhere, but since it is still a chat event, it could be catched by plugins(i.e. Prism-Bukkit) that log player commands. Further, the 2FA code should only be used once, so it's less sensitive than a password.

sgdc3:
>This indeed should not appear anywhere
We can just hide it from the server log, better than nothing

PR #1705, opened 2018-12-07
Sensitive commands [WIP]

sgdc3:
Test doesn't compile, needs feedback.
#1703 related


src/main/java/fr/xephi/authme/output/LogFilterService.java
```diff
@@ -0,0 +1,43 @@
...
+     * @return True if it is a sensitive AuthMe command, false otherwise
+     */
+    public boolean isSensitiveAuthMeCommand(String message) {
+        if (message == null || !message.contains(ISSUED_COMMAND_PREFIX_TEXT)) {
+            return false;
+        }
+        String rawCommand = message.substring(message.indexOf("/"));
+        List<String> components = Arrays.asList(rawCommand.split(" "));
+        FoundCommandResult command = commandMapper.mapPartsToCommand(null, components);
+        switch (command.getResultStatus()) {
```
ljacqu: This is really cool! Do you think we should somehow cache invocations of the commands? I'm a little worried about performance since this will be performed every time someone issues a command, and I'm not quite sure how efficient the commandMapper is. 
Failing that, I just saw that the LogFilterHelper has a test (`LogFilterHelperTest`) that ensures that all of those previously hard-coded labels are present. The alternative would be to keep the list and to update the test to use the sensitive flag.

src/main/java/fr/xephi/authme/output/LogFilterService.java
```diff
@@ -0,0 +1,43 @@
...
+            case UNKNOWN_LABEL:
+            case MISSING_BASE_COMMAND:
+                return false;
+            default:
+                break;
+        }
+        return command.getCommandDescription().isSensitive();
```
ljacqu: Minor: if we're playing code beauty contest I'd suggest putting the `return` statement in the `default` clause instead of `break`ing from it there; that way we oppose both cases directly to each other.

src/main/java/fr/xephi/authme/output/Log4JFilter.java
```diff
@@ -16,6 +16,12 @@
 
     private static final long serialVersionUID = -5594073755007974254L;
 
+    private LogFilterService filterService;
+
+    public Log4JFilter(LogFilterService filterService) {
+        this.filterService = filterService;
```
ljacqu: > I think that using the ConsoleFilter(LogFilterService) constructor is a bid dirty, any suggestion?

I don't really have a better suggestion. In theory we could put `@Inject` on the constructor and have a Log4JFilter instance injected into OnstartupTasks, but given the nature of how it is set up (in OnStartupTasks we have compatibility checks in case a certain log class doesn't exist) I wouldn't recommend doing that. I think your current solution works out nicely :)

src/main/java/fr/xephi/authme/output/LogFilterService.java
sgdc3: 1) yeah, performance can be a problem hmm... we should test it.
2) I'm not sure about the test :/ What would you do?

sgdc3: hmm, yeah, which portion of logic do you want to move?

sgdc3:
@ljacqu Could you please have again a look at this? :)

ljacqu:
Comment https://github.com/AuthMe/AuthMeReloaded/pull/1705#discussion_r240327001 still stands, ideally we can construct all combinations of sensitive commands and keep them in a Set, rather than calling the slow command mapper on every command. 

I'm also missing tests for the new LogFilterService ;)

PR #1706, opened 2018-12-07
WIP - Lowercase cleanup

sgdc3:
Tests needs to be adapted accordingly, any feedback?

### DIFF
--- a/src/main/java/fr/xephi/authme/command/executable/authme/RegisterAdminCommand.java
+++ b/src/main/java/fr/xephi/authme/command/executable/authme/RegisterAdminCommand.java
@@ -42,7 +42,6 @@ public class RegisterAdminCommand implements ExecutableCommand {
         // Get the player name and password
         final String playerName = arguments.get(0);
         final String playerPass = arguments.get(1);
-        final String playerNameLowerCase = playerName.toLowerCase();
 
         // Command logic
         ValidationResult passwordValidation = validationService.validatePassword(playerPass, playerName);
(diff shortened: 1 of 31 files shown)


ljacqu (commented):
I was skeptical when I saw this, but this actually constitutes a really nice clean up. Thanks man :)

src/main/java/fr/xephi/authme/command/executable/authme/RegisterAdminCommand.java
```diff
@@ -52,13 +51,13 @@ public void executeCommand(final CommandSender sender, List<String> arguments) {
...
+            if (dataSource.isAuthAvailable(playerName)) {
                 commonService.send(sender, MessageKey.NAME_ALREADY_REGISTERED);
                 return;
             }
-            HashedPassword hashedPassword = passwordSecurity.computeHash(playerPass, playerNameLowerCase);
+            HashedPassword hashedPassword = passwordSecurity.computeHash(playerPass, playerName);
             PlayerAuth auth = PlayerAuth.builder()
-                .name(playerNameLowerCase)
+                .name(playerName)
                 .realName(playerName)
```
ljacqu:
```java
     .name(playerName)
     .realName(playerName)
```

This has bit of a funny feel now. It's absolutely 100% correct. I'm just wondering if it'd make sense to have a single `name(String)` function, since obviously from the realname we immediately know the all-lowercase name? This is maybe something we can look at after this pull request (for the sake of merging this soon). I expect that in some cases we do only want to set the `name` only (without a realname), but in most cases it would be nice if we just had to pass the name once and could tell the PlayerAuth builder "hey you, do what you need to do" :smile:

sgdc3: yeah, we can enhance it in a future cleanup

src/main/java/fr/xephi/authme/data/captcha/CaptchaCodeStorage.java
```diff
@@ -76,10 +76,9 @@ private String generateCode(String name) {
      * @return true if the code matches, false otherwise
      */
     public boolean checkCode(String name, String code) {
-        String nameLowerCase = name.toLowerCase();
-        String savedCode = captchaCodes.get(nameLowerCase);
+        String savedCode = captchaCodes.get(name.toLowerCase());
         if (savedCode != null && savedCode.equalsIgnoreCase(code)) {
-            captchaCodes.remove(nameLowerCase);
+            captchaCodes.remove(name);
```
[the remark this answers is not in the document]
sgdc3: Good catch

sgdc3:
Now the problem is updating the unit test xD

PR #1721, opened 2018-12-31
Update the copyright year to 2018 just in time.

krusic22:
We can't just skip from 2017 to 2019.


sgdc3:
i'm late, as always xD

PR #1792, opened 2019-04-13
Translation to Serbian (RS)

Neeonn:


sgdc3:
Thank you! :)

ljacqu:
Hi @Neeonn , thanks a lot for the submission! I went ahead and changed the code from "rs" to "sr" so that it uses the [ISO-639-1 language code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) for Serbian. :) Cheers

Issue #1874, opened 2019-08-02
More granular logging settings


ljacqu:
We should add more debug messages in our code but with that we might potentially produce a lot of spam if we're only interested in detailed information in a certain area (e.g. permission handling). Goal of this issue is to introduce the ability to control what gets logged on which level on a more detailed level than just one configuration for the entire plugin.

sgdc3:
@ljacqu is anything left to do or we can close this issue?

ljacqu:
There‘s a lot left to do... We can‘t control log levels on class/package basis yet.

sgdc3:
@ljacqu oh lol, yeah, forgot about that part :P

ljacqu:
https://github.com/AuthMe/AuthMeReloaded/wiki/Configuring-logging will have to be updated once this is merged

PR #1875, opened 2019-08-02
#1874 Introduce individual ConsoleLogger instance per class

ljacqu:
- Create ConsoleLoggerFactory from which a separate logger can be created for each class
- Allows to support individual log level settings in the future

---

This doesn't change anything in behavior yet but now that each class has its own ConsoleLogger instance, we can introduce a more detailed setting for the log levels. The idea is that a user could then enable DEBUG logging just in a certain package of AuthMe, for example.

I ended up using an approach that is pretty much the same as Log4j or Slf4j use, namely to have a LoggerFactory where you pass in the class and based on that it creates a logger with that name. Initially I wanted to introduce a new annotation `@Logger` and have the injector just set those fields with a proper ConsoleLogger instance but it turned out to be more confusing than helpful. It would have required some extensive test changes to set it up in the tests as well and it could not have been used everywhere, seeing as we also have logging in enums, for example (see e.g. WalkFlySpeedRestoreType). So in any case some sort of factory-like approach would have been necessary and I didn't feel comfortable having two completely different ways of creating the logger in parallel. In the end it does require some boilerplate code, but I now understand why the other frameworks went like this, everything else seems to kind of get messy quite fast :sweat_smile:

### DIFF
--- a/src/test/java/fr/xephi/authme/data/limbo/persistence/LimboPersistenceTest.java
+++ b/src/test/java/fr/xephi/authme/data/limbo/persistence/LimboPersistenceTest.java
@@ -16,6 +16,8 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
 
+import java.util.logging.Logger;
+
 import static org.hamcrest.Matchers.both;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.equalTo;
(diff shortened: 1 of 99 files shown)


src/test/java/fr/xephi/authme/data/limbo/persistence/LimboPersistenceTest.java
```diff
@@ -122,7 +120,7 @@ public void shouldHandleExceptionWhenSavingLimbo() {
         // given
         Player player = mock(Player.class);
         LimboPlayer limbo = mock(LimboPlayer.class);
-        Logger logger = TestHelper.setupLogger();
+        java.util.logging.Logger logger = TestHelper.setupLogger();
```
ljacqu: woops

sgdc3:
I think we are ready for a merge, aren't we? @ljacqu :)

ljacqu:
I‘m happy to have it merged if you agree :) I didn‘t want to merge it "aggressively" since this is maybe a bit of a controversial change ^^

PR #1878, opened 2019-08-05
Add Maven profile to skip Javadoc generation for local builds

ljacqu:
Introduces a new Maven profile that allows to skip Javadoc generation for a fast local build (guess my machine is getting old because it really makes a difference for me :smile:)


ljacqu:
Awesome, thank you :)

Issue #1700, opened 2018-12-03
[bug] After authorization death in nether (teleportUnAuthedToSpawn: true)


Roper89:
@sgdc3 
If use `teleportUnAuthedToSpawn: true` — You will be killed if you leave the game from the worlds: "nether" and "the_end" 
Video [YouTube](https://www.youtube.com/watch?v=9mzHbGn3oA8)

_Minecraft: Latest Spigot 1.12.2
AuthMe-5.5.1-SNAPSHOT (#2181)_

sgdc3:
@Roper89 can you reproduce it with only authme as plugin?

ljacqu:
Perhaps related to #1624

Roper89:
Please fix it. 
If players are kicked out of the "nether" and "the_end" worlds, they die.

sgdc3:
Does it work with this jar? (Test with caution, backup your server FIRST!)
[AuthMe-5.5.1-SNAPSHOT.zip](https://github.com/AuthMe/AuthMeReloaded/files/2805315/AuthMe-5.5.1-SNAPSHOT.zip)

Roper89:
@sgdc3  This does not work.

sgdc3:
please share the configuration

Roper89:
https://pastebin.com/5zUBtgzs

sgdc3:
@Roper89 could you please try enabling debug in authme and send us the log? (it will be saved in the authme folder)

tzerber:
I also suffer from this and i can also attach logs / debug if that's going to help. My setup is VERY close to the setup from #1624 , latest spigot; latest essentials kit; worldedit ; worldguard ; bungeesuite (server is part of a proxy network).

tzerber:
I do found a fix, all of this is a config file issue. 
in config.yml make sure authme is the first in spawnPriority
`spawnPriority: authme,essentials,default,cmi,multiverse`
Also this should be enabled
`SaveQuitLocation: true`

Also edit your spawn.yml and doublecheck if the values there(map name / coords) are correct.

Roper89:
@sgdc3 , debug -- https://pastebin.com/2EsMXfK2

Roper89:
Deaths continue. Sorry, I did a test in a creative :(
If you set spawnPriority: authme
then the player will not be killed while the session lasts.
But if the session is out, the player will die when entering the game!

Roper89:
@sgdc3 @Hex3l  
[...]

PR #1880, opened 2019-08-06
Improves spawn choice if undefined

Hex3l:
#1700

### DIFF
--- a/src/main/java/fr/xephi/authme/settings/SpawnLoader.java
+++ b/src/main/java/fr/xephi/authme/settings/SpawnLoader.java
@@ -205,10 +205,33 @@ public class SpawnLoader implements Reloadable {
                 return spawnLoc;
             }
         }
+        if (!isValidSpawnPoint(world.getSpawnLocation())) {
+            for (World spawnWorld : Bukkit.getWorlds()) {
+                if (isValidSpawnPoint(spawnWorld.getSpawnLocation())) {
+                    world = spawnWorld;
+                    break;
+                }
+            }
+        }
         logger.debug("Fall back to default world spawn location. World: `{0}`", world.getName());
+        logger.warning("Seems like AuthMe is unable to find a proper spawn location. "
+            + "Set a location with the command '/authme setspawn'");
         return world.getSpawnLocation(); // return default location
     }
 
+    /**
+     * Checks if a given location is a valid spawn point [!= (0,0,0)].
+     *
+     * @param location The location to check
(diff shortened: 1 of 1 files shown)


ljacqu:
1) The for-loop `for (String priority : spawnPriority)` on L176 will by default always return a location so the loop you've added, I expect, will not be executed in many cases (for sure whenever the spawnPriority includes "default"). Seeing as your location check is quite lenient I would suggest to perform this check also before L202 when the spawnLocation is returned.

2) I think it would make sense to cache the world default spawn location. We shouldn't optimize for that case but I still think it might be worthwhile to have a field with the world that gets lazily initialized the first time we need it (the result of L211)

3) Minor: you could simplify the method to `return location.getX() != 0 || location.getY() != 0 || location.getZ() != 0` if you like, or just keep it as it is, of course. There's nothing wrong with it. :)

Hex3l:
The idea was to provide a working spawn system while AuthMe is not configured properly.

1. By default the configuration uses `spawnPriority: authme` , so the for-loop `for (String priority : spawnPriority)` will always trigger only the `authme` case. In this case we have two situations: 
    1. An authme spawn exists in the spawn.yml
        - In this case the user has manually set a spawn point so that spawn point will be used.
    2. There are no spawn points in the spawn.yml
        - In this case the for-loop ends and the `spawnLoc` will be null

   Since the spawn location is null the new validation for-loop will be triggered. 
An other way to handle this, as you suggested, is to place the validation for-loop at L182 and change the default configuration to something like `spawnPriority: authme,default`. Since this check refuses only the location (0,0,0) this new option could actually be a better implementation, but should also be noted a potential performance loss in extreme cases.

2. The idea was to cover the "not properly configured" case, that's why the warn should allert the end user to properly set a spawn location avoiding the hypothetical performance loss.

3. Thought it was more readable ^.^

I can't decide between the two implementation at point 1 because I think they have similar down sides so, I'm gonna rely on your choice 😄

ljacqu:
1. As far as I see in the code the default right now is "authme,essentials,cmi,multiverse,default". 

2. Totally agree, nothing to do here :)

Hex3l:
Well, seems like I had an old configuration 🤦‍♂
I'm going to move the verification + for-loop in the "default" case, sorry about that 😅

ljacqu:
Removing the release milestone since the issue this is for is already labeled as such -> keeps our bookkeeping free of duplicates

PR #1884, opened 2019-08-08
Listener optimizations

sgdc3:
Breaking changes:
- Dropped CraftBukkit support


src/test/java/fr/xephi/authme/listener/ListenerConsistencyTest.java
```diff
@@ -84,7 +77,7 @@ private static void checkCanceledAttribute(Class<?> listenerClass) {
         Method[] methods = listenerClass.getDeclaredMethods();
         for (Method method : methods) {
             if (isTestableMethod(method) && method.isAnnotationPresent(EventHandler.class)) {
-                if (CANCELED_EXCEPTIONS.contains(clazz + "#" + method.getName())) {
+                if (!method.getParameterTypes()[0].isAssignableFrom(Cancellable.class)) {
```
sgdc3: @ljacqu 😎

ljacqu: Hell yeah :smile::sparkles:

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -481,58 +446,67 @@ public void onSignChange(SignChangeEvent event) {
...
-    public void onPlayerShear(PlayerShearEntityEvent event) {
+    public void onPlayerDropItem(PlayerDropItemEvent event) {
         if (listenerService.shouldCancelEvent(event)) {
             event.setCancelled(true);
         }
     }
 
     @EventHandler(ignoreCancelled = true, priority = EventPriority.LOWEST)
-    public void onPlayerFish(PlayerFishEvent event) {
+    public void onPlayerHeldItem(PlayerItemHeldEvent event) {
```
ljacqu: Please add this method to PlayerListenerTest#shouldHandleSimpleCancelableEvents

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -229,175 +225,162 @@ public void onJoinMessage(PlayerJoinEvent event) {
...
     }
 
     @EventHandler(ignoreCancelled = true, priority = EventPriority.HIGHEST)
-    public void onPlayerKick(PlayerKickEvent event) {
-        // Note #831: Especially for offline CraftBukkit, we need to catch players being kicked because of
-        // "logged in from another location" and to cancel their kick
-        if (settings.getProperty(RestrictionSettings.FORCE_SINGLE_SESSION)
-            && event.getReason().contains("You logged in from another location")) {
-            event.setCancelled(true);
+    public void onPlayerRespawn(PlayerRespawnEvent event) {
```
ljacqu: This has no unit test :(

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -103,103 +98,104 @@
...
-        if (settings.getProperty(RestrictionSettings.ALLOW_COMMANDS).contains(cmd)) {
-            return;
-        }
-        final Player player = event.getPlayer();
-        if (!quickCommandsProtectionManager.isAllowed(player.getName())) {
-            event.setCancelled(true);
-            player.kickPlayer(messages.retrieveSingle(player, MessageKey.QUICK_COMMAND_PROTECTION_KICK));
+        final String name = event.getName();
+
+        // NOTE: getAddress() sometimes returning null, we don't want to handle this race condition
```
ljacqu: I really like this comment! This method would benefit of some tests, though :)  (only the try-catch stuff below is covered)

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -103,103 +98,104 @@
     @Inject
     private BungeeSender bungeeSender;
```
ljacqu: This field is now unused. The logger as well.
Also update the test please not to have bungee sender as mock field

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -481,58 +446,67 @@ public void onSignChange(SignChangeEvent event) {
...
+            /*
+             * @note little hack cause InventoryOpenEvent cannot be cancelled for
+             * real, cause no packet is sent to server by client for the main inv
+             */
+            bukkitService.scheduleSyncDelayedTask(player::closeInventory, 1);
         }
     }
 
+    @EventHandler(ignoreCancelled = true, priority = EventPriority.LOWEST)
+    public void onPlayerInventoryClick(InventoryClickEvent event) {
```
ljacqu: Just as feedback as a reviewer—unfortunately, I've had to check out the branch and reorder the methods the old way again so I could get a sensible diff. :) This would have benefited from two commits: one with actual method content changes and then a subsequent one reordering them (or the other way around, doesn't matter)

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -103,103 +98,104 @@
...
-        if (!settings.getProperty(RestrictionSettings.ALLOW_UNAUTHED_MOVEMENT)) {
-            // "cancel" the event
-            event.setTo(event.getFrom());
-            return;
+        // Slow, blocking checks
+        try {
+            final PlayerAuth auth = dataSource.getAuth(name);
+            final boolean isAuthAvailable = auth != null;
+            onJoinVerifier.checkKickNonRegistered(isAuthAvailable);
+            final JoiningPlayer joiningPlayer = JoiningPlayer.fromName(name);
```
ljacqu: Now that `JoiningPlayer#fromPlayer` is no longer used (except in a test) you can drop the class and all methods using it and just use the player name directly to look up permissions with.

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -229,175 +225,162 @@ public void onJoinMessage(PlayerJoinEvent event) {
...
-        final Player player = event.getPlayer();
-        final String name = player.getName();
+    /*
+     * Movement events
+     */
 
-        if (validationService.isUnrestricted(name)) {
+    @EventHandler(ignoreCancelled = true, priority = EventPriority.HIGHEST)
+    public void onPlayerMove(PlayerMoveEvent event) {
+        if (settings.getProperty(ALLOW_UNAUTHED_MOVEMENT) && settings.getProperty(ALLOWED_MOVEMENT_RADIUS) <= 0) {
```
ljacqu: On L353 the same `ALLOW_UNAUTHED_MOVEMENT` is checked if it's false; I would merge these two conditions at the start of this method like so:
```java
if (!settings.getProperty(ALLOW_UNAUTHED_MOVEMENT)) { // taken from L353 onwards
            // "cancel" the event
            event.setTo(event.getFrom());
            return;
} else if (settings.getProperty(ALLOWED_MOVEMENT_RADIUS) <= 0) {
  return; // unrestricted
}
```

src/main/java/fr/xephi/authme/message/MessageKey.java
```diff
@@ -146,6 +146,9 @@
     /** The server is full, try again later! */
     KICK_FULL_SERVER("on_join_validation.kick_full_server"),
 
+    /** An error occurred: unresolved player hostname! **/
+    KICK_UNRESOLVED_HOSTNAME("error.kick_unresolved_hostname"),
```
ljacqu: Please run the `verifyMessages` tool task so that all YML files will be updated with this new message :smiley:

src/test/java/fr/xephi/authme/listener/PlayerListenerTest.java
```diff
@@ -950,6 +986,13 @@ private static void verifyNoModifyingCalls(PlayerLoginEvent event) {
         verifyNoMoreInteractions(event);
     }
 
+    private static void verifyNoModifyingCalls(AsyncPlayerPreLoginEvent event) {
+        verify(event, atLeast(0)).getLoginResult();
+        verify(event, atLeast(0)).getAddress();
+        verify(event, atLeast(0)).getName();
+        verifyNoMoreInteractions(event);
```
ljacqu: Nice!

src/main/java/fr/xephi/authme/process/join/AsynchronousJoin.java
```diff
@@ -89,8 +89,9 @@ public void processJoin(final Player player) {
         }
 
         if (service.getProperty(RestrictionSettings.FORCE_SURVIVAL_MODE)
-            && !service.hasPermission(player, PlayerStatePermission.BYPASS_FORCE_SURVIVAL)) {
-            bukkitService.runTask(() -> player.setGameMode(GameMode.SURVIVAL));
+            && !service.hasPermission(player, PlayerStatePermission.BYPASS_FORCE_SURVIVAL)
+            && player.getGameMode() != GameMode.SURVIVAL) {
```
ljacqu: This is nice—if you want to micro optimize even more I'm guessing checking the game mode is cheaper than checking the permission :wink:

src/main/java/fr/xephi/authme/listener/EntityListener.java
```diff
@@ -81,6 +80,7 @@ public void entityRegainHealthEvent(EntityRegainHealthEvent event) {
         }
     }
 
+    //TODO sgdc3 20190808: Does it still make sense?
```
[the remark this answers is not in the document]
sgdc3: i don't know why we listen at the same event with 2 different priorities, the only thing i can think of is that we want to prevent another plugin from un-cancelling the event :/

src/main/java/fr/xephi/authme/listener/EntityListener.java
```diff
@@ -97,11 +97,8 @@ public void onLowestEntityInteract(EntityInteractEvent event) {
 
     @EventHandler(ignoreCancelled = true, priority = EventPriority.LOWEST)
     public void onProjectileLaunch(ProjectileLaunchEvent event) {
-        if (event.getEntity() == null) {
-            return;
-        }
+        final Projectile projectile = event.getEntity();
 
-        Projectile projectile = event.getEntity();
         // In the Bukkit API prior to 1.7, getShooter() returns a LivingEntity instead of a ProjectileSource
```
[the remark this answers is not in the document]
sgdc3: Nice catch!

src/main/java/fr/xephi/authme/listener/PlayerListener.java
sgdc3: Sorry! xD

src/main/java/fr/xephi/authme/listener/PlayerListener.java
sgdc3: Not applicable, the movement logic wouldn't make sense.

src/main/java/fr/xephi/authme/message/MessageKey.java
sgdc3: uh the tool seems to be broken, it will place "null" instead of the missing translation values

src/main/java/fr/xephi/authme/listener/PlayerListener.java
ljacqu: How does moving a check up in the method affect the movement logic?

src/main/java/fr/xephi/authme/message/MessageKey.java
ljacqu: Fixed on master, thanks for the hint :smile:

src/main/java/fr/xephi/authme/listener/PlayerListener.java
sgdc3: We want to allow movement on the Y axis

src/main/java/fr/xephi/authme/listener/PlayerListener.java
sgdc3: It is so simple i have no idea how to test it in a way that makes sense xD

src/main/java/fr/xephi/authme/listener/PlayerListener.java
sgdc3: Hmm, could you please help me creating this test?

src/main/java/fr/xephi/authme/listener/PlayerListener.java
ljacqu: Ah sorry, you're right :relaxed:

src/main/java/fr/xephi/authme/listener/PlayerListener.java
ljacqu: It's a few tests:
1) No interaction with event if NO_TELEPORT is set to true
2) No interaction with event if listenerService.shouldCancelEvent is `false`
3) Finally, should set the spawn location to the event if we indeed get this far
4) Bonus: test that ensures everything is OK if the spawnLoader does not return a location ^^

src/main/java/fr/xephi/authme/listener/PlayerListener.java
ljacqu: Of course! If you look at PlayerListener#shouldPerformAllJoinVerificationsSuccessfullyPreLoginLowest you can basically just copy it BUT create the `preLoginEvent` to have `null` as IP address. Also because a message is taken from the messages service, mock it to return something, i.e. under "given" add `given(messages.retrieveSingle(name, MessageKey.KICK_UNRESOLVED_HOSTNAME)).willReturn("Unresolved hostname");`

Afterwards under "then" you can `verify(event).disallow(AsyncPlayerPreLoginEvent.Result.KICK_OTHER, "Unresolved hostname")` because that's exactly what you expect to have been called on the event. I would also add a `verifyNoInteractions(onJoinVerifier)` so the test makes sure we don't do anything with it in this case

PR #1898, opened 2019-08-24
Update messages_br.yml

brunoxkk0:
Added messages who was missing in this translation.


ljacqu:
Thank you :smile::sparkles:

PR #1913, opened 2019-09-17
Fix unregistered players teleport to spawn with unforced registration

Kongolan:
Considering that I posted the same bug 4 years back (which was fixed on an old version), you should really consider that there are some servers, which do not force registration for players... since the whole test class is written only with forced registration in mind.

If you set teleportUnAuthToSpawn to true and also forceRegistration to false, unregistered players always end up at spawn, while they should not be altered by AuthMe and login like vanilla. Once you register that player it works perfectly fine.

### DIFF
--- a/src/main/java/fr/xephi/authme/service/TeleportationService.java
+++ b/src/main/java/fr/xephi/authme/service/TeleportationService.java
@@ -13,6 +13,7 @@ import fr.xephi.authme.initialization.Reloadable;
 import fr.xephi.authme.output.ConsoleLoggerFactory;
 import fr.xephi.authme.settings.Settings;
 import fr.xephi.authme.settings.SpawnLoader;
+import fr.xephi.authme.settings.properties.RegistrationSettings;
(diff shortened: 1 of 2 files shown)


ljacqu:
Looks fine to me.
@sgdc3 – I remember in the past you were worried about checking if an auth is available too frequently. Maybe we should think about issuing a warning if registration is optional but the data source is not cached? 
@Kongolan – thanks for the contribution. You mention that there are no tests that test for optional registration. Feel free to add some if you like.

Kongolan:
So far I only added the tests for optional registration to make it compile. So I did not actually write a solid test code. But it's really something which is missed out all the time.

Kongolan:
Looking into this again since i was going to check for the 1.15 update... What is the codeclimate issue which needs to be fixed? Is that the reason why the pull request is not applied yet?

sgdc3:
Hmm dataSource.isAuthAvailable is a blocking call, are we shure that the teleport method is never called from the main thread?

PR #1921, opened 2019-10-04
Updated messages_nl.yml with missing translations.

jordieh:
I also added two notes for the next translator about difficult to translate messages.

Should we translate `captcha` to `puzzel` (which means `puzzle` in English)? `captcha` is not really a well-known word in Dutch.

### DIFF
--- a/src/main/resources/messages/messages_nl.yml
+++ b/src/main/resources/messages/messages_nl.yml
@@ -40,8 +40,7 @@ error:
     max_registration: 'Je hebt het maximum aantal registraties overschreden (%reg_count/%max_acc: %reg_names).'
     logged_in: '&cJe bent al ingelogd!'
     kick_for_vip: '&cEen VIP-gebruiker heeft ingelogd toen de server vol was!'
-    # TODO (Translator): Het onderste bericht is lastig te vertalen, moeten we 'computernaam' of 'hostname' gebruiken? Welke structuur kunnen wij het beste gebruiken voor 'unresolved'?
-    kick_unresolved_hostname: '&cEr heeft een fout plaatsgevonden: computernaam van de speler kon niet worden opgehaald!'
+    kick_unresolved_hostname: '&cEr heeft een fout plaatsgevonden: hostname van de speler kon niet gevonden worden!'
     tempban_max_logins: '&cJe bent tijdelijk gebanned omdat het inloggen te vaak mislukt is.'
 
 # AntiBot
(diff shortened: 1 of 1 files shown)


src/main/resources/messages/messages_nl.yml
```diff
@@ -40,7 +40,8 @@ error:
     max_registration: 'Je hebt het maximum aantal registraties overschreden (%reg_count/%max_acc: %reg_names).'
     logged_in: '&cJe bent al ingelogd!'
     kick_for_vip: '&cEen VIP-gebruiker heeft ingelogd toen de server vol was!'
-    # TODO kick_unresolved_hostname: '&cAn error occurred: unresolved player hostname!'
+    # TODO (Translator): Het onderste bericht is lastig te vertalen, moeten we 'computernaam' of 'hostname' gebruiken? Welke structuur kunnen wij het beste gebruiken voor 'unresolved'?
+    kick_unresolved_hostname: '&cEr heeft een fout plaatsgevonden: computernaam van de speler kon niet worden opgehaald!'
```
ljacqu: Ik zou kiezen voor hostname. This is about not being able to resolve something from the user's IP address; computer name sounds like the name you can choose for your computer in the settings. But I'm not a native Dutch speaker.
opgehaald -> maybe opgelost? At least in German I would say something along the lines of "de hostnaam van de ip address kan niet worden opgelost"

src/main/resources/messages/messages_nl.yml
```diff
@@ -119,18 +120,19 @@ captcha:
     usage_captcha: '&3Om in te loggen moet je een captcha-code oplossen, gebruik het commando: /captcha %captcha_code'
     wrong_captcha: '&cVerkeerde captcha-code, typ alsjeblieft "/captcha %captcha_code" in de chat!'
     valid_captcha: '&2De captcha-code is geldig!'
-    # TODO captcha_for_registration: 'To register you have to solve a captcha first, please use the command: /captcha %captcha_code'
-    # TODO register_captcha_valid: '&2Valid captcha! You may now register with /register'
+    # TODO (Translator): Moeten wij het woord 'captcha' laten staan? Of is puzzel een betere verwoording?
+    captcha_for_registration: 'Om je te registreren moet je eerst een puzzel oplossen. Gebruik alsjeblieft het volgende commando: /captcha %captcha_code'
+    register_captcha_valid: 'Valide puzzel! Je mag je nu registreren met /register'
```
ljacqu: Again, not a native speaker, but if you do see "captcha" in NL-language sites I would vote for that as it's more specific than _puzzel_. I see Wikipedia uses (captcha)[https://nl.wikipedia.org/wiki/Captcha]. But consistency is more important, IMHO—you can choose which one is more appropriate and we can look that all the messages will use the same terminology.

ljacqu:
Thanks for doing this! :smile: I added some comments that would be good to address but if you'd like we can already merge now and see the rest later. Please just note that the comments will be removed at some point because we have an automatic job that adds new missing messages and it doesn't preserve any comments.

src/main/resources/messages/messages_nl.yml
jordieh: I changed it to hostname, as that is used by KPN (A major Dutch internet provider). https://www.internedservices.nl/faq/wat-vul-ik-in-bij-hostname/ . Furthermore, i changed `kon niet worden opgehaald!` to `kon niet gevonden worden!`, which litterally translates to `could not be found` instead of `could not be collected/picked up`. I think this is the best translation possible.

Ps. Het is `het ip adres` in plaats van `de ip address` :D

src/main/resources/messages/messages_nl.yml
jordieh: I changed it to captcha. I could only find one reference to captcha. Maybe this should be revisited later but for now it's the best i can make of it.

jordieh:
I removed the comments and adjusted the translations according to your comments. I think this is ready to be merged :)!

ljacqu:
Great, thanks a lot for everything! :smile:

src/main/resources/messages/messages_nl.yml
ljacqu: Thanks for the correction :smile: I should have known both the spelling and the article... especially the spelling :D I need to use Dutch more again~

Issue #1448, opened 2017-12-08
API: Authme email integration


SuperJogosDBZ:
I want to put a plugin that forces the player to register his email on the server, however to do this I need to see and check registered email, can you tell me how can I do this?

ljacqu:
This is covered by our email registration case.

In your config.yml:
```yaml
    registration:
[...]
        force: true
        # Type of registration: PASSWORD or EMAIL
        # PASSWORD = account is registered with a password supplied by the user;
        # EMAIL = password is generated and sent to the email provided by the user.
        # More info at https://github.com/AuthMe/AuthMeReloaded/wiki/Registration
        type: 'EMAIL'
```

SuperJogosDBZ:
This is not what I wanted, I would like those who are already registered to register the email and not those who will still register. So I need a method to check if the game has registered email.

ljacqu:
> So I need a method to check if the [player] has registered email.

Are you developing your own plugin that wants to retrieve AuthMe data, or what exactly are you trying to do?

SuperJogosDBZ:
Exactly!

ljacqu:
Thanks for the info. This relates to #1373 (just mentioning it so the issues get linked)

Issue #1373, opened 2017-10-22
API: Get player information


DevNatan:
Well I'm making a plugin that it should show some information that are in AuthMe, like password, email ... But I do not find any method in the API that shows the email for example.

I searched the AuthMe code and found "PlayerAuth" and "PlayerCache" only they have an annotation "@Inject" which I do not know for what it is, I just wanted to have a way of invoking "PlayerCache" and get the information of the players, returning in a "PlayerAuth".

My authme version: 5.3.2 (b1726)

ljacqu:
When we add this I'd like whatever new methods we add to return an interface, that PlayerAuth can implement (or some other class), but I'd just like our PlayerAuth class not to be directly coupled with API methods.

DevNatan:
I understand, but is not there any way to get a player's email through your name or unique id?

ljacqu:
At the moment, there isn't.

DevNatan:
I urgently ask YouTube to add methods that can do such a function of obtaining information. Be it not necessarily encrypted. It would greatly increase the number of plugins that can use your API.

Sorry for persistence, but i'm Brazilian, and we're known for It... 😁😂😋

ljacqu:
This was solved within #1448, so marking as a duplicate for our bookkeeping purposes (although this is the original issue and it's more general)

PR #1978, opened 2019-12-29
Get player email through API

YellowZaki:
Relates to https://github.com/AuthMe/AuthMeReloaded/issues/1373 https://github.com/AuthMe/AuthMeReloaded/issues/1448

### DIFF
--- a/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
+++ b/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
@@ -365,7 +365,11 @@ public class AuthMeApi {
             auth = dataSource.getAuth(playerName);
         }
         if (auth != null) {
-            return auth.getEmail();
+            String email = auth.getEmail();
+            if (email == null || email.equals("your@email.com")) {
+                return null;
+            }
+            return email;
         }
         return null;
     }    
@@ -377,7 +381,7 @@ public class AuthMeApi {
      * @return true if the player has linked email
      */
     public boolean hasEmail(String playerName) {
-        return (getEmail(playerName) != null && !getEmail(playerName).equals("your@email.com"));
+        return (getEmail(playerName) != null);
     } 
     
 }
--- a/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
+++ b/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
@@ -366,7 +366,7 @@ public class AuthMeApi {
         }
         if (auth != null) {
             String email = auth.getEmail();
-            if (email == null || email.equals("your@email.com")) {
+            if (("your@email.com").equals(email)) {
                 return null;
             }
             return email;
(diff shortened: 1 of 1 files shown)


YellowZaki:
Well, I didn't realised default email is your@email.com, not null.

ljacqu:
Sorry for the late reply.
I think it would make sense to null out the email as well in the above method if it's the default. Isn't there a constant on PlayerAuth that has the value `your@email.com`?

In the future I'd like to go in the direction of having some sort of interface we'll publicly expose with the information (name / email / etc.) so that a user will just get some sort of ApiUser object and doesn't have to call multiple methods that load the PlayerAuth each time to get basic info. But I'm happy to have these methods for now since I don't think that will happen soon... :P But with that in mind, maybe it doesn't make sense to have the `hasEmail` method.

YellowZaki:
There is no constant for your@email.com.
I have made that getEmail() returns null if default.
I think it is good to keep hasEmail(), it is more user friendly.

ljacqu:
PlayerAuth#DB_EMAIL_DEFAULT 

The thing about the other method is, yes, it's a nice readable method, but it doesn't add new functionality. I'd be on board with such a method if we had a dedicated class to users or something, but since we anyway intend on providing an alternative to the getEmail method, I'd like to keep the number of methods to a minimum that a user has to be familiar with. Again, only because we intend to provide another approach later on and because we only have one API class.

src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
```diff
@@ -352,4 +352,36 @@ public String getCountryCode(String ip) {
...
+     * @return The email of the player
+     */
+    public String getEmail(String playerName) {
+        PlayerAuth auth = playerCache.getAuth(playerName);
+        if (auth == null) {
+            auth = dataSource.getAuth(playerName);
+        }
+        if (auth != null) {
+            String email = auth.getEmail();
+            if (("your@email.com").equals(email)) {
```
sgdc3: PlayerAuth#DB_EMAIL_DEFAULT

src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
```diff
@@ -352,4 +352,36 @@ public String getCountryCode(String ip) {
...
+    }    
+    
+    /**
+     * Check if player has linked email
+     *
+     * @param playerName The name of the player to check
+     * @return true if the player has linked email
+     */
+    public boolean hasEmail(String playerName) {
+        return (getEmail(playerName) != null);
```
sgdc3: Don't wrap it with ()

YellowZaki:
Since https://github.com/AuthMe/AuthMeReloaded/pull/2000 has been implemented, I close this

ljacqu:
Sorry for the lack of update, and thanks for your contribution :) This is what sparked the addition of the player API :)

Issue #1979, opened 2019-12-31
No more GeoLite2 databases updates since December 30, 2019 available


chmatse:
### What behaviour is observed:
[04:34:13] [Craft Scheduler Thread - 1/INFO]: [AuthMe] Downloading GEO IP database, because the old database is older than 30 days or doesn't exist
[04:34:13] [Craft Scheduler Thread - 1/WARN]: [AuthMe] Could not download GeoLiteAPI database [UnknownHostException]: geolite.maxmind.com

### What behaviour is expected:
The download of the latest GEO IP databases.

### Steps/models to reproduce:
According to:
https://dev.maxmind.com/geoip/geoip2/geolite2/#Databases
https://blog.maxmind.com/2019/12/18/significant-changes-to-accessing-and-using-geolite2-databases/

Starting December 30, 2019, we will be requiring users of our GeoLite2 databases to register for a MaxMind account and obtain a license key in order to download GeoLite2 databases. We will continue to offer the GeoLite2 databases without charge, and with the ability to redistribute with proper attribution and in compliance with privacy regulations. In addition, we are introducing a new end-user license agreement to govern your use of the GeoLite2 databases. Previously, GeoLite2 databases were accessible for download to the public on our developer website and were licensed under the Creative Commons Attribution-ShareAlike 4.0 International License.

Starting December 30, 2019, downloads will no longer be served from our public GeoLite2 page, from geolite.maxmind.com/download/geoip/database/*, or from any other public URL. See the section below for steps on how to migrate to the new download mechanism.

### AuthMe build number:
https://ci.codemc.io/job/AuthMe/job/AuthMeReloaded/lastSuccessfulBuild/artifact/target/AuthMe-5.6.0-SNAPSHOT.jar

MaximeMichaud:
I agree.
Need to fix or find an alternative.

ljacqu:
Thanks for the bump. You‘re right. They now require a signup and that a key be sent along with the download... As a quick fix we could sign up and just provide the same key for all AuthMe users, until MaxMind shuts us down – a quick search showed that other projects just store the databases elsewhere and download from there. We need to discuss this internally and will fix it then.

FabioZumbi12:
Essentials is using that keys and maded a wiki to help users to get a key and use it, may help you all: https://essentialsx.cf/wiki/GeoIP.html

(i already created a key on maxmind site and its working fine on essentials)

bonjourkemal:
[...]

PR #1990, opened 2020-01-20
Update the GeoIp database download method [BREAKING]

sgdc3:
Now GeoIp database updates require a ClientID and a LicenseKey, which can be obtained for free at https://www.maxmind.com/en/accounts/current/license-key

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1,5 +1,5 @@
 <!-- AUTO-GENERATED FILE! Do not edit this directly -->
-<!-- File auto-generated on Thu Oct 17 08:29:25 CEST 2019. See docs/config/config.tpl.md -->
+<!-- File auto-generated on Mon Jan 20 14:16:50 CET 2020. See docs/config/config.tpl.md -->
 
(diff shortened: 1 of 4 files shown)


sgdc3:
#1979

Xephi:
Maybe we can let users choose between downloading the whole database or using the api directly ? 
https://github.com/maxmind/GeoIP2-java

ljacqu:
Great job! I have two questions about error scenarios:
1. if one of the config values is empty, could we log something meaningful? (Pointing them to the config.yml, maybe to the properties‘ path)
2. if the download somehow fails, the plugin should still work and start up normally, right?

sgdc3:
@Xephi the api isn't as fast as the local db, i personally had issues with antibot plugins using the rest api, due to rate limits and the performance impact

sgdc3:
@ljacqu 
1) Ok, there is already a message but it doesn't point to the config path
2) Yes, the geoip check will just be ignored, and a warning message will be printed on startup

PR #1993, opened 2020-01-21
Add method to get the registration date to the API

Drc-DEV:
No API method was written when the feature was introduced, there are some use-cases where this data is needed.

### DIFF
--- a/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
+++ b/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
@@ -197,6 +197,32 @@ public class AuthMeApi {
     }
 
     /**
+     * Get the registration (AuthMe) timestamp of a player.
+     *
+     * @param playerName The name of the player to process
+     *
+     * @return The timestamp of when the player was registered, or null if the player doesn't exist or is not registered
+     */
+    public Instant getRegistrationTime(String playerName) {
+        Long registrationDate = getRegistrationMillis(playerName);
+        return registrationDate == null ? null : Instant.ofEpochMilli(registrationDate);
+    }
+
+    private Long getRegistrationMillis(String playerName) {
+        if (!isRegistered(playerName.toLowerCase())) {
+            return null;
+        }
+        PlayerAuth auth = playerCache.getAuth(playerName);
+        if (auth == null) {
+            auth = dataSource.getAuth(playerName);
+        }
+        if (auth != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
```diff
@@ -196,6 +196,32 @@ private Long getLastLoginMillis(String playerName) {
...
+     *
+     * @return The timestamp of when the player was registered, or null if the player doesn't exist or is not registered
+     */
+    public Instant getRegistrationTime(String playerName) {
+        Long registrationDate = getRegistrationMillis(playerName);
+        return registrationDate == null ? null : Instant.ofEpochMilli(registrationDate);
+    }
+
+    private Long getRegistrationMillis(String playerName) {
+        if (!isRegistered(playerName.toLowerCase())) {
```
sgdc3: I think that this check can be avoided as we already check if auth != null below

ljacqu:
At this point I think instead of merging this and PR #1978 we should actually do the long-term solution and introduce an ApiUser interface (or whatever we want to call it) that allows to get a _few_ of the interesting values in one API method. I can try to do this maybe this weekend.

Drc-DEV:
> 
> 
> At this point I think instead of merging this and PR #1978 we should actually do the long-term solution and introduce an ApiUser interface (or whatever we want to call it) that allows to get a _few_ of the interesting values in one API method. I can try to do this maybe this weekend.

That would be defintely better, but I'd still add these methods to the current API because they are really essential, and I think that you should still keep the the v3 API for backwards compatibility anyways.

src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
Drc-DEV: Done

ljacqu:
Definitely, I'm talking about adding that one method onto the current API :)

sgdc3:
In the meantime i'll merge this, as we have no eta for the UserApi

ljacqu:
I said this weekend?

ljacqu:
Basically @Drc-DEV count on this being reverted. Sorry for the arbitrary behavior on our side.

Drc-DEV:
> 
> 
> Basically @Drc-DEV count on this being reverted. Sorry for the arbitrary behavior on our side.

Just make it deprecated if you make the UserApi. No need to remove a couple of lines of code.

PR #2000, opened 2020-01-26
#1448 Create AuthMePlayer to get player data from API with

ljacqu:
Introduces an API method to get a lot of player info with one load from the database. Unit tests will still follow but I open this PR early to get some feedback on the naming (and anything else). Ideally we won't have to remove or rename anything in the future since this is an API so it's really worth looking at our API methods carefully.

For the naming of _AuthMePlayer_: "ApiPlayer" would make sense for us in _our_ codebase but not for anyone that is integrating our API. I also chose "Player" over "User" as I guess that's really the more used terminology in Minecraft (you have _players_ on the server; it's _players_ who register themselves in AuthMe).

cc: @RikoDEV @YellowZaki

### DIFF
--- a/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
+++ b/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
@@ -133,20 +133,17 @@ public class AuthMeApi {
     }
 
     /**
-     * Get the registration ip address of a player.
+     * Returns the AuthMe info of the given player's name, or null if the player doesn't exist.
      *
-     * @param playerName The name of the player to process
-     * @return The registration ip address of the player
+     * @param playerName The player name to look up
+     * @return AuthMe player info, or null if not available
      */
-    public String getRegistrationIp(String playerName) {
+    public AuthMePlayer getPlayerInfo(String playerName) {
         PlayerAuth auth = playerCache.getAuth(playerName);
         if (auth == null) {
             auth = dataSource.getAuth(playerName);
         }
-        if (auth != null) {
-            return auth.getRegistrationIp();
-        }
-        return null;
+        return AuthMePlayerImpl.fromPlayerAuth(auth);
     }
 
     /**
@@ -180,7 +177,6 @@ public class AuthMeApi {
      * Get the last (AuthMe) login date of a player.
      *
      * @param playerName The name of the player to process
-     *
      * @return The date of the last login, or null if the player doesn't exist or has never logged in
      * @deprecated Use Java 8's Instant method {@link #getLastLoginTime(String)}
      */
@@ -209,29 +205,6 @@ public class AuthMeApi {
         }
         if (auth != null) {
             return auth.getLastLogin();
-        }
-        return null;
-    }
-
-    /**
-     * Get the registration (AuthMe) timestamp of a player.
-     *
-     * @param playerName The name of the player to process
-     *
-     * @return The timestamp of when the player was registered, or null if the player doesn't exist or is not registered
-     */
-    public Instant getRegistrationTime(String playerName) {
-        Long registrationDate = getRegistrationMillis(playerName);
-        return registrationDate == null ? null : Instant.ofEpochMilli(registrationDate);
-    }
-
-    private Long getRegistrationMillis(String playerName) {
-        PlayerAuth auth = playerCache.getAuth(playerName);
-        if (auth == null) {
-            auth = dataSource.getAuth(playerName);
-        }
-        if (auth != null) {
-            return auth.getRegistrationDate();
         }
         return null;
     }
--- a/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
+++ b/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
@@ -21,6 +21,7 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
+import java.util.Optional;
 
 /**
  * The current API of AuthMe.
@@ -133,12 +134,12 @@ public class AuthMeApi {
     }
 
     /**
-     * Returns the AuthMe info of the given player's name, or null if the player doesn't exist.
+     * Returns the AuthMe info of the given player's name, or empty optional if the player doesn't exist.
      *
      * @param playerName The player name to look up
-     * @return AuthMe player info, or null if not available
+     * @return AuthMe player info, or empty optional if the player doesn't exist
      */
-    public AuthMePlayer getPlayerInfo(String playerName) {
+    public Optional<AuthMePlayer> getPlayerInfo(String playerName) {
         PlayerAuth auth = playerCache.getAuth(playerName);
         if (auth == null) {
             auth = dataSource.getAuth(playerName);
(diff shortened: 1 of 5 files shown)


src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
```diff
@@ -133,20 +133,17 @@ public Location getLastLocation(Player player) {
...
-     * Get the registration ip address of a player.
+     * Returns the AuthMe info of the given player's name, or null if the player doesn't exist.
      *
-     * @param playerName The name of the player to process
-     * @return The registration ip address of the player
+     * @param playerName The player name to look up
+     * @return AuthMe player info, or null if not available
      */
-    public String getRegistrationIp(String playerName) {
+    public AuthMePlayer getPlayerInfo(String playerName) {
```
sgdc3: Api v4, deprecate api v3

sgdc3: Optional

src/main/java/fr/xephi/authme/api/v3/AuthMePlayer.java
```diff
@@ -0,0 +1,62 @@
...
+    String getName();
+
+    /**
+     * Returns the UUID of the player as given by the server (may be offline UUID or not).
+     * The UUID is null if AuthMe is configured not to store the UUID or if the data is not
+     * present (e.g. older record).
+     *
+     * @return player uuid, or null if not available
+     */
+    UUID getUuid();
```
sgdc3: Optional

src/main/java/fr/xephi/authme/api/v3/AuthMePlayer.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Returns the email address associated with this player, or null if not available.
+     *
+     * @return player's email or null
+     */
+    String getEmail();
```
sgdc3: Optional

src/main/java/fr/xephi/authme/api/v3/AuthMePlayer.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * @return the registration date of the player's account - never null
+     */
+    Instant getRegistrationDate();
```
sgdc3: Optional

src/main/java/fr/xephi/authme/api/v3/AuthMePlayer.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Returns the IP address with which the player's account was registered. May be null
+     * for older accounts, or if the account was registered by someone else (e.g. by an admin).
+     *
+     * @return the ip address used during the registration of the account, or null
+     */
+    String getRegistrationIpAddress();
```
sgdc3: Optional

src/main/java/fr/xephi/authme/api/v3/AuthMePlayer.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Returns the last login date of the player. May be null if the player never logged in.
+     *
+     * @return date the player last logged in successfully, or null if not applicable
+     */
+    Instant getLastLoginDate();
```
sgdc3: Optional

src/main/java/fr/xephi/authme/api/v3/AuthMePlayer.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Returns the IP address the player last logged in with successfully. May be null if the
+     * player never logged in.
+     *
+     * @return ip address the player last logged in with successfully, or null if not applicable
+     */
+    String getLastLoginIpAddress();
```
sgdc3: Optional

src/main/java/fr/xephi/authme/api/v3/AuthMePlayer.java
```diff
@@ -0,0 +1,62 @@
+package fr.xephi.authme.api.v3;
+
+import java.time.Instant;
+import java.util.UUID;
+
+/**
+ * Read-only player info exposed in the AuthMe API.
```
games647: With the current implementation, we should maybe also mention that this object is a copy that won't update after the API method is called. So it could be outdated later without any notification.

src/main/java/fr/xephi/authme/api/v3/AuthMePlayerImpl.java
```diff
@@ -0,0 +1,92 @@
...
+        return registrationIpAddress;
+    }
+
+    @Override
+    public Instant getLastLoginDate() {
+        return lastLoginDate;
+    }
+
+    @Override
+    public String getLastLoginIpAddress() {
```
games647: Is there a reason we don't use InetAddresses? I know they perform a DNS request, but only on DNS names not IP addresses.

src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
ljacqu: > Api v4, deprecate api v3

No.
This is just a method addition and shouldn't warrant a new API version. Unless you mean the removed method, but that's one that's two days old and wasn't part of any release (even beta).

src/main/java/fr/xephi/authme/api/v3/AuthMePlayerImpl.java
ljacqu: Most cases I'm thinking of would compare the IP address with another string IP from elsewhere, or store it somewhere, no? I'm worried it will be a more annoying type than helpful as I'd anticipate a lot of uses having to unwrap it. But if you think otherwise I'm happy to change it.

(Also I'm lazy because constructing InetAddresses might throw UnknownHostException, which should never happen, but we'd still have to handle them properly :laughing:). It's not an excuse to not do it and as mentioned, I'm happy to do it if you really want.

Another alternative is offering both `String` and `InetAddress` but I think we should be careful with our interface(s), i.e. choose an approach and kind of stick with it. Say, if we added more fields to the AuthMePlayer it would quickly become confusing.

src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
sgdc3: Fair, but i would like to change that in the future, with proper versioning, maybe by splitting the api into a separate maven module.

ljacqu:
Happy to use `Optional`s (can't reply to @games647 comment directly). Has the advantage that the user is _forced_ to handle the "not present" state. Personally not a huge fan but for APIs it's definitely nice. Just note that it will not be in sync with the other existing methods.

src/main/java/fr/xephi/authme/api/v3/AuthMePlayer.java
ljacqu: One comment would have been enough :P

src/main/java/fr/xephi/authme/api/v3/AuthMePlayerImpl.java
games647: >Also I'm lazy because constructing InetAddresses might throw UnknownHostException,

Well if it's correct IP address we could ignore it. If not, it doesn't make sense to return this invalid string either.

>Most cases I'm thinking of would compare the IP address with another string IP from elsewhere, or store it somewhere, no? I'm worried it will be a more annoying type than helpful as I'd anticipate a lot of uses having to unwrap it. But if you think otherwise I'm happy to change it.

Ok I agree with you. I think the methods of `InetAddress` are most likely not relevant over the `String` use cases.

PR #2005, opened 2020-02-03
Queue async operations on a per-player basis [WIP]

sgdc3:


sgdc3:
Depends on #1992

src/main/java/fr/xephi/authme/process/AsyncUserScheduler.java
```diff
@@ -0,0 +1,92 @@
...
+        private Map<K, SequentialExecutor> executors;
+        private Function<Runnable, BukkitTask> scheduler;
+
+        public KeyedSequentialScheduler(Function<Runnable, BukkitTask> scheduler) {
+            this.executors = new LinkedHashMap<>();
+            this.scheduler = scheduler;
+        }
+
+        public void submit(K key, Runnable runnable) {
+            executors.computeIfAbsent(key, k -> new SequentialExecutor(scheduler, () -> executors.remove(key)))
```
games647: Potential memory leak

src/main/java/fr/xephi/authme/process/AsyncUserScheduler.java
```diff
@@ -0,0 +1,92 @@
...
+        private BukkitTask executor;
+
+        public SequentialExecutor(Function<Runnable, BukkitTask> scheduler, Runnable callback) {
+            this.queue = new LinkedBlockingQueue<>();
+            this.scheduler = scheduler;
+            this.callback = callback;
+        }
+
+        public void submit(Runnable task) {
+            queue.add(task);
```
games647: This doesn't seem thread-safe. Considering two calls to the `submit` method. Then both tasks will be added, but `scheduler.apply()` will be invoked twice with two new tasks on the Bukkit scheduler resulting to two new threads. 

T1 and T2 are the threads:
T1: 1 iteration of the while loop
-- 1 task remaining
T1: evaluate `while(!queue.isEmpty())` -> `true` - enter `while` loop
T2: 1 iteration of the while loop
-- `queue.poll` wasn't executed yet by T1 and so `queue` is still not empty
-- T2 now removes the last task
T1: `queue.poll()` - `NoSuchElementExeception`

TuxCoding (changes requested):
I kind of understand what you are doing here. You are trying to perform asynchronous tasks sequential on per player basis. I guess with the goal to reduce conflicts between them? Do you also thought about locks? This allows more flexibility like read/write differentiation, protecting only thread-unsafe operations meanwhile other parts could execute in `concurrently` and in `parallel`.

sgdc3:
@games647 i think that both read/write operations needs to be sequential, I can't think about any situation where locks might be better. Can you make some examples?

src/main/java/fr/xephi/authme/process/AsyncUserScheduler.java
sgdc3: Hmm I see, which pattern should we use instead?

games647: The easiest solution would be to lock on the queue on `isEmpty` + `poll` operations.

PR #2080, opened 2020-04-19
Unrestricted inventories fix

DevLeoko:
UNRESTRICTED_INVENTORIES is a lowercase string set, so the inventory title also needs to be converted to lowercase in order to be able to match the inventory names specified in the config


sgdc3:
Thank you

PR #2104, opened 2020-05-09
forceLogin now doesn't skip 2FA system

### LINKED ISSUE #2103: [API] forceLogin method skips TOTP authentication
### What behaviour is observed:
When using external plugins, such as FastLogin, they only have access to the forceLogin method, which skips TOTP entirely.

### What behaviour is expected:
This method should only bypass a player's password login, but never the 2FA method. It makes it completely useless!

### AuthMe build number:
Build 2385

JotaDevMC:
Fixes #2103


ljacqu:
+1 but would maybe be worth having a second optional argument to the command to optionally skip it as well? I can imagine of some use cases where it would be useful. 

Also what is that `TODO` comment in the code you've added?

PR #2138, opened 2020-06-24
Update Euskara (basque) translation

galaipa:
I translated the plugin in the year 2015, so there were many new missing translations. It should be 100% again.


sgdc3:
Thank you!

PR #2188, opened 2020-08-28
Fix build

garagepoort:
Overload reload method to pass in a boolean to turn on/off the migration check.
Old method is still there.

### DIFF
--- a/src/main/java/fr/xephi/authme/datasource/AbstractSqlDataSource.java
+++ b/src/main/java/fr/xephi/authme/datasource/AbstractSqlDataSource.java
@@ -167,4 +167,6 @@ public abstract class AbstractSqlDataSource implements DataSource {
             return DataSourceValueImpl.unknownRow();
         }
     }
+
+    public abstract void reload(boolean migrate);
 }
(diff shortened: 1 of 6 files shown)


ljacqu (changes requested):
The failing test uses an old SQLite database, which has to be migrated (now for two reasons). I'm fairly sure the test fails because the migration code has issues. Fixing SqLiteMigraterIntegrationTest to not migrate anymore kind of beats the purpose

ljacqu:
Fixed in https://github.com/AuthMe/AuthMeReloaded/commit/e92721e59741b6caf81fbe9c7612118a68f84251, thanks for raising awareness to the issue

garagepoort:
I didn't remove the migration. I removed triggering the migration infinite times...

ljacqu:
> I removed triggering the migration infinite times...

That's exactly my point: the test was failing because AuthMe (in the real code) would also trigger a migration infinite times. The issue was with the migration _itself_. 

>  Is it because the check now doesn't always return true after migration ?

Yes, the commit message explains why that was the case before.

PR #2231, opened 2020-11-25
Queue async operations on a per-player basis (v2)

sgdc3:
See comments on #2005

### DIFF
--- a/src/main/java/fr/xephi/authme/process/AsyncUserScheduler.java
+++ b/src/main/java/fr/xephi/authme/process/AsyncUserScheduler.java
@@ -1,15 +1,10 @@
 package fr.xephi.authme.process;
 
+import com.jano7.executor.KeySequentialRunner;
 import fr.xephi.authme.service.BukkitService;
 import org.bukkit.entity.Player;
-import org.bukkit.scheduler.BukkitTask;
 
 import javax.inject.Inject;
-import java.util.LinkedHashMap;
-import java.util.Map;
-import java.util.Queue;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.function.Function;
 
 /**
  * Handles the queue of async tasks on a per-player basis.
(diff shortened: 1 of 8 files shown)


src/main/java/fr/xephi/authme/process/AsyncUserScheduler.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public class AsyncUserScheduler {
+
+    @Inject
+    private BukkitService bukkitService;
+
+    private KeySequentialRunner<String> asyncUserScheduler;
+
+    AsyncUserScheduler() {
+        this.asyncUserScheduler = new KeySequentialRunner<>(command -> bukkitService.runTaskAsynchronously(command));
```
games647, line 20: Maybe I'm missing here something, but this appears just submits a task to be run async and returns immediately. The tasks will still run in concurrently.

TuxCoding (changes requested):

src/main/java/fr/xephi/authme/process/AsyncUserScheduler.java
sgdc3: Hmm, what do you mean exactly? I implemented a custom Executor which schedules async tasks using the bukkit scheduler.

games647: I thought the idea is to run the player specific code sequential for the same player. `runTaskAsynchronously` submits the task to Bukkit and then returns. This means the queue for this `UserScheduler` is empty. Another task can be submitted while the older is still running, because of the empty queue.

sgdc3: Oh, I'm soo stupid lol, guess we have to use a custom executor at this point.

sgdc3: Something isn't quite right anyway, this is the library I used: https://github.com/jano7/executor
From what I understand it won't wait for the task to complete, right?

games647: The ReadMe sounds like exactly what you want. Processing for the equal key is performed sequentially. 

I might just realized that I'm completely wrong. We are already creating a custom executor, because of the lambda call. So in theory there should be no issue.

sgdc3: yep but what does track when a task is completed?

sgdc3: Ok figured it out lol, any keyed runner keeps a runnable queue and runs tasks sequentially. Works as expected :)

games647: > yep but what does track when a task is completed?

They seem to wrap the runnable to track it.

PR #2262, opened 2021-01-15
Specify the Driver Class Name ("com.mysql.cj.jdbc.Driver")

MacacoLew:
TODO: Configurable Driver Class Name

### DIFF
--- a/src/main/java/fr/xephi/authme/datasource/MySQL.java
+++ b/src/main/java/fr/xephi/authme/datasource/MySQL.java
@@ -43,6 +43,7 @@ public class MySQL extends AbstractSqlDataSource {
     private String port;
     private String username;
     private String password;
+    private String className;
     private String database;
     private String tableName;
     private int poolSize;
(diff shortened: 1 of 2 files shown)


ljacqu:
Fine by me if we can make it configurable (we can give you pointers if you need), and would be interesting if you could share the background to this addition?

MacacoLew:
> Fine by me if we can make it configurable (we can give you pointers if you need), and would be interesting if you could share the background to this addition?

I will update this PR to make the class name configurable soon, and i maked this PR because AuthMe in MYSQL Mode stop the server because doens't recognize the mysql server. (Latest MySQL library in my fork, and this PR fixs it, tested)

MacacoLew:
> Fine by me if we can make it configurable (we can give you pointers if you need), and would be interesting if you could share the background to this addition?

U like the changes? say me if im wrong with some typos if u can :), im spanish and my english is not good too.

sgdc3:
Looks good, thank you :)

PR #2275, opened 2021-02-03
If a player is not registered then we send a LOGIN plugin message on join

bundabrg:
Presently when AuthMeBungee is used and requireAuth is set to true a player needs to login before doing anything.

This will allow players who are not registered (have no password) to be treated as if they are logged in by broadcasting the LOGIN message.

This means that players who are registered will be required to login and will be unable to execute any command either on the server or on the proxy, and those who are not registered will still be able to play without needing to register.

### DIFF
--- a/src/main/java/fr/xephi/authme/process/Management.java
+++ b/src/main/java/fr/xephi/authme/process/Management.java
@@ -57,6 +57,10 @@ public class Management {
         runTask(() -> asynchronousLogin.forceLogin(player));
     }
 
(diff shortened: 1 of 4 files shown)


sgdc3:
https://github.com/AuthMe/AuthMeBungee/issues/54

sgdc3:
This PR also removes duplicated login messages when a login is performed due to a valid session being detected. This is achieved by implementing a "silent" argument in the forceLogin method.

PR #2276, opened 2021-02-03
When performing a login we will try for up to 30s in case the player is not yet fully logged in

### LINKED ISSUE #2126: Autologin - bungeecord
When i login on our first server "survival" and i teleport to second server "skyblock" it wants sometiems login again, when i log in, it says to me that i am already logged, but i cant move i can only use jump for moving so i have to reconnect the server to be able to move properly on the server.

When i log in server "skyblock" and i teleport to "survival" i have to always log in again.
When i login server "skyblock" or "survival" and i teleport to our third server "hard" i have to always login again aswell.

Its just weird that sometimes i dont need to login again when i teleport to "skyblock server" but i have to login always again when i teleport to the servers "survival" and "hard".

I have autologin set to true in config on bungeecord server. We are using MYSQL.
We have the latest build 2402 on spigot servers and the latest build 158 on bungeecord server.

digors
[config_bungee.txt](https://github.com/AuthMe/AuthMeReloaded/files/4732929/config_bungee.txt)
[config_hard.txt](https://github.com/AuthMe/AuthMeReloaded/files/4732931/config_hard.txt)
[config_skyblock.txt](https://github.com/AuthMe/AuthMeReloaded/files/4732932/config_skyblock.txt)
[config_survival.txt](https://github.com/AuthMe/AuthMeReloaded/files/4732933/config_survival.txt)

bundabrg:
Sometimes when switching between bungee servers the player is not yet fully connected before the backend server receives the PERFORM_LOGIN packet. This means the player is null and they will require another login attempt.

We schedule a task that will try every second for 30 seconds to attempt the login, delaying the first attempt by 5 ticks. This should fix this issue.

Closes #2126

### DIFF
--- a/src/main/java/fr/xephi/authme/service/bungeecord/BungeeReceiver.java
+++ b/src/main/java/fr/xephi/authme/service/bungeecord/BungeeReceiver.java
@@ -147,10 +147,14 @@ public class BungeeReceiver implements PluginMessageListener, SettingsDependent 
         }
     }
 
+    /**
+     * Attempt to log in player for up to 30s
+     *
+     * @param name username of player
+     */
     private void performLogin(final String name) {
-        // Attempt the login for up to 30s in case a slow player login causes this to fail
         bukkitService.runTaskTimerAsynchronously(new BukkitRunnable() {
-            int count=0;
+            private int count = 0;
 
             @Override
             public void run() {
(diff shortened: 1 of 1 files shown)


sgdc3:
Closing in favour of #2285

PR #2290, opened 2021-02-17
Addition of /log command in the allowCommands sequence

Eric-01:
AuthMeReloaded already registers the `/log` command by default. Then this command cannot be used **after** the user logs in (which of course is good), but the same command cannot be used **before** the user logs in as well, which makes this particular command completely useless (by default of course, without modifying the config file).

My commit resolves this situation by adding the `/log` command to the default `allowCommands` sequence so that from now on this command can be used alongside the `/login` and `/l` commands.


sgdc3:
Looks good, thank you

PR #2302, opened 2021-03-13
Fix #2018. using player's health value instead of isDead (unreliable)

### LINKED ISSUE #2018: Player returns to death location when leaving server after dying
**Bug description:**
When you die by any cause, and then you go to "Title screen" on the death screen that shows up, then join in again into the server, and succesfully log in, you literally revive in the same place. 

**My words to say:**
I think its a bug because I don't find any node on the config file related to this. Already tested it on an empty server and it stills happening (I'm using the lastest version). This is a highly abusable bug, hope it gets fixed soon! Thanks for reading.
EDIT: The best way to solve this in my opinion, would be to save the death screen event and get it back to him after he succesfully logs in.

BRA1L0R:
Technical information available in the last comments of the issue,  here's a recap:

When Spigot (in my case Tuinity) calls the PlayerQuitEvent, it returns a Player instance which is flagged as both dead and offline. When AuthMeReloaded saves the quit location it checks for those two values, which return garbage data. A quick and easy way to check if the Player was dead at the moment of the quit is to check it's health, which is exactly what I did. The health value is fortunately reliable and shows 0.0 if the player was dead on exit

The exploit that could take place before this fix:
If you selected "Go to title screen" upon death and rejoined the server you could "bypass" your respawn location and directly spawn on the location of your death. I've tested my fix and it works!

### DIFF
--- a/src/main/java/fr/xephi/authme/settings/SpawnLoader.java
+++ b/src/main/java/fr/xephi/authme/settings/SpawnLoader.java
@@ -273,7 +273,7 @@ public class SpawnLoader implements Reloadable {
      * @return location of the given player if alive, spawn location if dead.
      */
     public Location getPlayerLocationOrSpawn(Player player) {
-        if (player.isOnline() && player.isDead()) {
+        if (player.getHealth() <= 0.0) {
             return getSpawnLocation(player);
         }
         return player.getLocation();


BRA1L0R:
Players do not respawn at their bed, need to fix

BRA1L0R:
I'm reopening the pull request because it's still better than the current status of AMR + it could still work if essentials is in charge of managing bed spawns / world spawn points.

ljacqu (approved):

Hex3l (approved):

Issue #2112, opened 2020-05-13
Disabled cache lets run SQL commands on the main thread


games647:
**Background:**
For inventory protection, we are checking every time the inventory is changed if the user is registered ([If not logged in](https://github.com/AuthMe/AuthMeReloaded/blob/3c9b1077f2fe776339e7899a582d2e718050427d/src/main/java/fr/xephi/authme/listener/protocollib/InventoryPacketAdapter.java#L86)). In #2109 we found out that if caching is disabled this runs directly a database query on the main thread. Despite this being bad practice, this runs for every packet, so possible a lot of times.

**Solution:**
This part is found at numerous places like event listeners, so the easy solution to just run it asynchronous is useless. In the configuration we recommend disable the cache for BungeeCord and website registration. 

However in case of BungeeCord we already request refreshes on something like password changes to other servers, still currently not registering. For website registration we could periodically refresh the data or use something like Redis. We already do the first one every 5 minutes. The best solution would be to have single task that refreshes it depending on a last updated timestamp for all players.

What is your recommendation? Do I miss here something?

7wOv6ySCjo:
Sorry for reply this. So if i have only one BungeeCord environment and only one auth server (preLobby) can leave 'caching: true' ?

games647:
>Sorry for reply this. 

No need to. This is open for everyone.

>So if i have only one BungeeCord environment and only one auth server (preLobby) can leave 'caching: true' ?

We recommend disabling the cache for always fetching the newest data. In this case, you **theoretically** can enable caching, because the newest data is already locally and it's the only source for modifications.

Issue #1830, opened 2019-06-07
Even though, it's duplicated. Unregistered unforced players should have their Inventory shown...


Kongolan:
Version: Build #2256 (Tue Jun 04)

You can "reproduce" it by

1. Turn off the force register. (under # Only registered and logged in players can play.)
2. Turn on the inventory protection. (ProtectInventoryBeforeLogIn: false)
3. Log in with an not registered account.

Players without registration can't access their inventory. The reason for this, is that the ProtocolLib Adapter doesn't account unregistered players and force register FALSE properly.

Please check this out, since I checked into it, but i can't see where the ProtocolLib inventory protection is called, since it's not in the PlayerListener nor in the AsynchronousJoin.

```
        if (isAuthAvailable) {
            // Protect inventory
            if (service.getProperty(PROTECT_INVENTORY_BEFORE_LOGIN)) {
                ProtectInventoryEvent ev = bukkitService.createAndCallEvent(
                    isAsync -> new ProtectInventoryEvent(player, isAsync));
                if (ev.isCancelled()) {
                    player.updateInventory();
                    ConsoleLogger.fine("ProtectInventoryEvent has been cancelled for " + player.getName() + "...");
                }
            }
```
This account it properly, but the event call is strange for me. It seems like the ProtocolLib inventory protection is handled independent of that.

This is a duplicate of:
https://github.com/AuthMe/AuthMeReloaded/issues/1443
https://github.com/AuthMe/AuthMeReloaded/issues/1752

Kongolan:
Using 
```
    public void register(BukkitService bukkitService, boolean hideNow) {
        ProtocolLibrary.getProtocolManager().addPacketListener(this);

        if (hideNow) {
			bukkitService.getOnlinePlayers().stream().filter(p -> playerCache.isAuthenticated(p.getName())
					|| (!service.getProperty(RegistrationSettings.FORCE) && !database.isAuthAvailable(p.getName())))
					.forEach(this::sendBlankInventoryPacket);
        }
    }
```
in InventoryPacketAdapter.java did somehow not work. I have no insight into how the inventory adapter is handled but without ProtocolLib this problems do not appear.

ljacqu:
[...]

Issue #1752, opened 2019-02-09
If "ProtectInventoryBeforeLogIn:" is set to true before a player registers, after they register and login their inventory doesn't get restored


axemwa:
If "ProtectInventoryBeforeLogIn:" is set to true before a player registers, after they register and login their inventory doesn't get restored. It only restores their inventory on login if they were already registered before the setting was set to true

AuthMe version 5.5.1-SNAPSHOT-b2202

sgdc3:
hmm i think this is by design, but we can improve this

axemwa:
Yeah it's not so bad for new servers but servers that have already been active and have had active players no so much since their inventory and original position before registration doesn't get restored.
I'm pretty sure this is the same case as when a player unregisters and goes to register again

Kongolan:
There is another bug, which refers to this. If you don't force registration, and have ProtectInventoryBeforeLogin on, then you don't see your inventory until you register and login. So basically the ProtocolLib adapter doesn't check for unregistered players, which it should account.

ljacqu:
Hi @axemwa , to the best of my abilities I've tried to change the behavior to make more sense. You can check it with the latest build on https://ci.codemc.org/job/AuthMe/job/AuthMeReloaded/ I'd be happy for some feedback as there are a lot of cases to consider and you may have different expectations :)

axemwa:
seems to be workin fine now tvym

PR #2358, opened 2021-06-18
Check restriction status only on cache

TuxCoding:
TL;DR Registration status will only be checked if the data source is cached to prevent blocking SQL queries on the main thread. Another solution would be a plugin wide notification system when we query the player status, but it's more complex. See more details below at **Refresh events**.

@ljacqu, I tried to preserve your behavior from [this comment](https://github.com/AuthMe/AuthMeReloaded/issues/1830#issuecomment-504699971). Therefore the proposed changes will still show the inventory for unregistered players, however only if we use the local cache.

Commit desc:

In #2112, we found out that in configurations with disabled cache (like Bungee), this adapter will perform blocking queries on the main thread which impacts the performance.

The original code (81cf14fbc16aaecf7ac72dfea9a8acbff21f4f6d) is in place to
preserve the inventory for unregistered and configurations without enforcing a registration (#1830, #1752). Alternatives are:

1. Send the inventory on registration like p.updateInventory()
  * Still hides it before for unregistered
2. Checking for the enforce registration setting would mean it hides also the inventory for unregistered players
3. Get a notification on player status changes
  * Requires a complex setup to propagate the changes across spigot servers
  or different solution for web interfaces
  * Refresh events, when the status is updated within the plugin itself would be possible, however requires previous queries like registration/login requests. Instant reports about registration will still be complicated. Example: Every time we make queries, save the result locally and fire an event for our components to check for potential changes without making the same query again. Nevertheless this again delays changes if there is no need to that.

So the best solution is to use the cache if available and if not fallback to safe defaults.

### DIFF
--- a/src/main/java/fr/xephi/authme/data/auth/PlayerCache.java
+++ b/src/main/java/fr/xephi/authme/data/auth/PlayerCache.java
@@ -10,6 +10,7 @@ import java.util.concurrent.ConcurrentHashMap;
 public class PlayerCache {
 
     private final Map<String, PlayerAuth> cache = new ConcurrentHashMap<>();
+    private final Map<String, RegistrationStatus> registeredCache = new ConcurrentHashMap<>();
 
     PlayerCache() {
     }
@@ -20,6 +21,7 @@ public class PlayerCache {
      * @param auth the player auth object to save
      */
     public void updatePlayer(PlayerAuth auth) {
+        registeredCache.put(auth.getNickname().toLowerCase(), RegistrationStatus.REGISTERED);
         cache.put(auth.getNickname().toLowerCase(), auth);
     }
 
@@ -30,6 +32,7 @@ public class PlayerCache {
      */
     public void removePlayer(String user) {
         cache.remove(user.toLowerCase());
+        registeredCache.remove(user.toLowerCase());
     }
 
     /**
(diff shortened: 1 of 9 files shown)


games647:
Fixes #2112

ljacqu (approved):

ljacqu:
Very pragmatic, looks great to me! Thanks

games647:
Okay it's not that easy. I forgot the following case:

DataSource not cached, registration not enforced:
Currently: It would always hide the inventory, because we don't know the registration status
Issue: We need to know the status, unregistered are allowed to do anything meanwhile registered needs to be protected.

games647:
Since the `ListenerService` is also affected from this issue, I changed to behavior now to use caches there too. Currently the result will be cached on join and updated after login. Only the restriction settings will be using the cache, other components like the register command will bypass it in order to fetch the newest data.

Downside is that changes from other sources like websites won't be available until the player joins the server again. An alternative could be cache every `isAuthAvailable` if the player is online to have an up to date result when the plugin requests any updates from the database. Furthermore periodic updates like our existing cache (every 5min) would be also possible.

sgdc3 (approved):

sgdc3:
Looks good to me

games647:
I think I should add some unit tests too, because this is critical area. Just to be sure ;)

sgdc3:
@games647 hello, any news on this?

games647:
Argh, I was too busy with other stuff. I'll try on Thursday.

sgdc3:
@games647 re-bump? 😄

games647:
Test are ready. However I encountered a race condition. When preventing the inventory data to be sent the status of the registration status isn't loaded yet. Therefore we should always prevent it at this stage. My issue is creating a good design. Currently we fire a `ProtectInventoryEvent` only if necessary on player join. My idea would be to always fire it on join and fire `RestoreInventoryEvent` once data is loaded.

sgdc3:
@games647 @ljacqu bump

sgdc3:
@games647 @ljacqu re-bump? 😄

sgdc3:
@games647 hello? 😅

Issue #1782, opened 2019-03-26
ipv6 ip of user doesn't go into MySQL database


vophamthanhan:
error while trying to store ipv6 address for authenticated user into a MySQL database.
```
[AuthMe] Error during SQL operation: [MysqlDataTruncation]: Data truncation: Data too long for column 'ip' at row 1
```

for example, a user may join with an ipv6 address of `[d82e:b902:18db:2bf3:9f59:cfda:75af:8584]`, compared to an example ipv4 address of `54.175.226.192`, the address is different in size and formatting.

Versions:
AuthMeReloaded v5.4.0 (build: 1877)
mysql  Ver 14.14 Distrib 5.7.25, for Linux (x86_64) using  EditLine wrapper

ITZVGcGPmO:
from a security point of view, keep in mind that ipv6 addresses have a prefix and a suffix.
the prefix is given out by the ISP, while the suffix can be changed at will by device.
if an IP addresses is used to limit(say registration), i suggest only the prefix of ipv6 be used.
the prefix can be fetched by [some regex](https://regex101.com/r/6a1KG4/1).

there are multiple different [ipv6 representations](https://en.wikipedia.org/wiki/IPv6_address#Representation) to go along.

Xephi:
I think the problem is we try to add '[' and ']' characters onto database too (maybe? we need to do some tests about that)
The ip columns current type are all VARCHAR(40), since ipv4 are 15 characters long (max) and ipv6 are 39 characters long (max)

Take a look into : 
https://github.com/AuthMe/AuthMeReloaded/blob/master/src/main/java/fr/xephi/authme/datasource/MySQL.java#L193
https://github.com/AuthMe/AuthMeReloaded/blob/master/src/main/java/fr/xephi/authme/datasource/MySQL.java#L211

By the way, it seems we can have a problem when the type is ipv4 mapped ipv6, we could have : `0000:0000:0000:0000:0000:ffff:255.255.255.255` which could be 45 characters long (max), i'll take a look if we try to store that kind of ip and check if we can do a fix for this specific case

Thanks for the report, we'll update this issue when we have some news

ITZVGcGPmO:
i've got some new flashy regex.

the output of the [this filter](https://regex101.com/r/6a1KG4/5) can be considered the "real ip", and can be stored in sql. (ipv4-6 mappings and [brackets])
[...]

PR #2440, opened 2021-10-24
Allow REGISTRATION_IP and LAST_IP to store IPv6 addresses

cofob:
Fixing #1782 by increasing the maximum `VARCHAR` field length. You should also add a migration to update to a new length from previous versions.


sgdc3:
A migration should be implemented before merging

cofob:
> A migration should be implemented before merging

Yes, but I'm noob in Java. What files are responsible for the migration?

Issue #1827, opened 2019-06-04
Velocity proxy support suggestion.


jokbon:
Velocity proxy support suggestion.
Curently teleporting to other server after loging in doesnt work with auth me and velocity.

Wats that:
https://www.velocitypowered.com/
https://discord.gg/8cB9Bgf

sgdc3:
Temp fix for basic features: https://github.com/VelocityPowered/BungeeQuack

endigma:
Any updates on this functionality?

tommasobenatti:
Any update?

OpenSource03:
Please do not ignore this. It's pretty important for a lot of users.

andrsn0w:
yes

Dagrond:
Velocity is used by more and more people, we need this.

endigma:
cursed dead issue

andrsn0w:
up

compieter-gh:
Bump

Why isn't this included by default? 
Kinda been using a outdated build from
https://ci.hex3l.net/job/AuthMereloaded/

4drian3d:
Currently there are some external plugins that allow AuthMe to be used correctly under a Velocity network, e.g. [AuthMeVelocity](https://github.com/4drian3d/AuthMeVelocity)

PR #2442, opened 2021-10-27
Add option to change plugin messaging channel in configuration.

Rollczi:
Linked: https://github.com/AuthMe/AuthMeReloaded/issues/1827

### DIFF
--- a/src/main/java/fr/xephi/authme/service/BukkitService.java
+++ b/src/main/java/fr/xephi/authme/service/BukkitService.java
@@ -14,9 +14,7 @@ import org.bukkit.command.CommandSender;
 import org.bukkit.configuration.file.YamlConfiguration;
 import org.bukkit.entity.Player;
 import org.bukkit.event.Event;
-import org.bukkit.plugin.Plugin;
 import org.bukkit.scheduler.BukkitRunnable;
-import org.bukkit.scheduler.BukkitScheduler;
 import org.bukkit.scheduler.BukkitTask;
 
 import javax.inject.Inject;
@@ -303,12 +301,13 @@ public class BukkitService implements SettingsDependent {
     /**
      * Send the specified message to bungeecord using the first available player connection.
      *
+     * @param channel the channel
      * @param bytes the message
      */
-    public void sendBungeeMessage(byte[] bytes) {
+    public void sendBungeeMessage(String channel, byte[] bytes) {
         Player player = Iterables.getFirst(getOnlinePlayers(), null);
         if (player != null) {
-            player.sendPluginMessage(authMe, "BungeeCord", bytes);
+            player.sendPluginMessage(authMe, channel, bytes);
         }
     }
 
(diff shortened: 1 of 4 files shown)


sgdc3:
The main concern with this option is that when a channel different from the "BungeeCord" one is used there is nothing preventing clients to inject messages and bypass the login request. (BungeeCord proxy ensures that the client can't send messages on the "BungeeCord" channel)

sgdc3:
The only possible solution I can think of is to force the server owner to set a shared secret when the used channel is not the default "BungeeCord" one. Idk, it's just an idea.

Rollczi:
@sgdc3 
https://github.com/LiteDevelopers/AuthMeLiteVelocity/blob/0d93d8f3f1ce388115fcc0678294da99efac9956/src/main/java/fr/xephi/authmevelocity/listeners/VelocityMessageListener.java#L46
this check should solve the problem

a similar solution can be found in fastlogin plugin for velocity:
https://github.com/games647/FastLogin/blob/d2c94af4a776b93c52fa97fd5636aa3924ca7af3/velocity/src/main/java/com/github/games647/fastlogin/velocity/listener/PluginMessageListener.java#L72

I guess that should prevent this from happening, if not, then maybe it will try to solve it with a secret key.

Hex3l:
> @sgdc3 https://github.com/LiteDevelopers/AuthMeLiteVelocity/blob/0d93d8f3f1ce388115fcc0678294da99efac9956/src/main/java/fr/xephi/authmevelocity/listeners/VelocityMessageListener.java#L46 this check should solve the problem
> 
> a similar solution can be found in fastlogin plugin for velocity: https://github.com/games647/FastLogin/blob/d2c94af4a776b93c52fa97fd5636aa3924ca7af3/velocity/src/main/java/com/github/games647/fastlogin/velocity/listener/PluginMessageListener.java#L72
> 
> I guess that should prevent this from happening, if not, then maybe it will try to solve it with a secret key.

I gave a quick look at provided solutions, if I am not mistaken the problem @sgdc3 is referring to is not verifying if a message is coming from a server (both the links refer to a proxy env recieving a message from a server) but veryifing if it is coming **from** the proxy while beeing a server.
I have zero knowledge on velocity but if a channel doesn't provide a local setting, as BungeeCord channel does, an user will be able to send a plugin message emulating a [_proxy login request_](https://github.com/AuthMe/AuthMeReloaded/blob/b74496fb1529726ca120a1691d96e381e784ab7e/src/main/java/fr/xephi/authme/service/bungeecord/BungeeReceiver.java#L128) that we take as valid. A solution, as @sgdc3 already stated, would be to validate a secret [(as done in FastLogin for example)](https://github.com/games647/FastLogin/blob/15fee9293739b89f602a7b6ec3cee1cc8580edd9/core/src/main/java/com/github/games647/fastlogin/core/message/LoginActionMessage.java#L83) or finding a local setting in velocity channel system.

Issue #2503, opened 2022-02-11
Armor stand spectating login


MithrandirCraft:
### What feature do you want to see added?

Forcing players which are either logging in or registring to spectate an invisible armorstand while logging in for 0 position and head pitch yaw movement.

Making the armorstand's XYZ EulerAngle's configurable, as well as actually using this login approach or not.

The suggestion was made by @krusic22 on the AuthMe Discord.

I tried implementing this myself for a later PR, but found Authme's arquitecture quite daunting, and couldn't make my code fit propperly. So I am now suggesting a more experienced dev to implement this instead.
Here's my initial approach to a class that would set up the Armor Stand for one: https://gist.github.com/MithrandirCraft/a0f2ec81c9e1ef0a9c95e962a66bd487

### Are there any alternatives?

Yeah. AuthMe used to have a rather rudimentary head position rollback (I do not know if this is still in place), and also still has a position rollback, both of which are not too good at preventing both types of movements, but do indeed roll them back (the first of which being head movement, I do not know if is still present in configuration).

PR #2552, opened 2022-06-03
SpectateLogin implementation

mrchefran:
Implementation of the idea proposed in #2503

Feature description: puts the player in an invisible armorstand for the time of authorization and fixes the direction of the view. After authorization is completed, the stand disappears

### DIFF
--- a/src/main/java/fr/xephi/authme/listener/PlayerListener.java
+++ b/src/main/java/fr/xephi/authme/listener/PlayerListener.java
@@ -11,6 +11,7 @@ import fr.xephi.authme.process.Management;
 import fr.xephi.authme.service.AntiBotService;
 import fr.xephi.authme.service.BukkitService;
 import fr.xephi.authme.service.JoinMessageService;
+import fr.xephi.authme.service.SpectateLoginService;
 import fr.xephi.authme.service.TeleportationService;
 import fr.xephi.authme.service.ValidationService;
 import fr.xephi.authme.settings.Settings;
@@ -19,6 +20,7 @@ import fr.xephi.authme.settings.properties.HooksSettings;
 import fr.xephi.authme.settings.properties.RegistrationSettings;
 import fr.xephi.authme.settings.properties.RestrictionSettings;
 import org.bukkit.ChatColor;
+import org.bukkit.GameMode;
 import org.bukkit.Location;
 import org.bukkit.entity.HumanEntity;
 import org.bukkit.entity.Player;
@@ -49,6 +51,8 @@ import org.bukkit.event.player.PlayerPickupItemEvent;
 import org.bukkit.event.player.PlayerQuitEvent;
 import org.bukkit.event.player.PlayerRespawnEvent;
 import org.bukkit.event.player.PlayerShearEntityEvent;
+import org.bukkit.event.player.PlayerTeleportEvent;
+import org.bukkit.event.player.PlayerToggleSneakEvent;
 import org.bukkit.inventory.InventoryView;
 
 import javax.inject.Inject;
@@ -90,6 +94,8 @@ public class PlayerListener implements Listener {
     private PermissionsManager permissionsManager;
     @Inject
     private QuickCommandsProtectionManager quickCommandsProtectionManager;
+    @Inject
+    private SpectateLoginService spectateLoginService;
 
     // Lowest priority to apply fast protection checks
     @EventHandler(priority = EventPriority.LOWEST)
@@ -505,6 +511,27 @@ public class PlayerListener implements Listener {
     public void onPlayerInventoryClick(InventoryClickEvent event) {
         if (listenerService.shouldCancelEvent(event.getWhoClicked())
             && !isInventoryWhitelisted(event.getView())) {
+            event.setCancelled(true);
+        }
+    }
+
+    @EventHandler(ignoreCancelled = true, priority = EventPriority.LOWEST)
+    public void onToggleSneak(PlayerToggleSneakEvent event) {
+        if (listenerService.shouldCancelEvent(event.getPlayer())
+            && (settings.getProperty(RestrictionSettings.SPECTATE_STAND_LOGIN)
+            || spectateLoginService.hasStand(event.getPlayer()))) {
+            event.setCancelled(true);
+        }
+    }
+
+    @EventHandler(ignoreCancelled = true, priority = EventPriority.HIGHEST)
+    public void onTeleport(PlayerTeleportEvent event) {
+        if (listenerService.shouldCancelEvent(event.getPlayer())
+            && event.getCause() == PlayerTeleportEvent.TeleportCause.SPECTATE
+            && event.getPlayer().getGameMode() == GameMode.SPECTATOR
+            && (settings.getProperty(RestrictionSettings.SPECTATE_STAND_LOGIN)
+            || spectateLoginService.hasStand(event.getPlayer()))) {
+            spectateLoginService.updateTarget(event.getPlayer());
             event.setCancelled(true);
         }
     }
--- a/src/main/java/fr/xephi/authme/process/join/AsynchronousJoin.java
+++ b/src/main/java/fr/xephi/authme/process/join/AsynchronousJoin.java
@@ -14,6 +14,7 @@ import fr.xephi.authme.service.BukkitService;
 import fr.xephi.authme.service.CommonService;
 import fr.xephi.authme.service.PluginHookService;
 import fr.xephi.authme.service.SessionService;
+import fr.xephi.authme.service.SpectateLoginService;
 import fr.xephi.authme.service.ValidationService;
 import fr.xephi.authme.service.bungeecord.BungeeSender;
 import fr.xephi.authme.service.bungeecord.MessageType;
@@ -80,6 +81,9 @@ public class AsynchronousJoin implements AsynchronousProcess {
 
     @Inject
     private ProxySessionManager proxySessionManager;
+
+    @Inject
+    private SpectateLoginService spectateLoginService;
 
     AsynchronousJoin() {
     }
@@ -191,6 +195,14 @@ public class AsynchronousJoin implements AsynchronousProcess {
                 int blindTimeOut = (registrationTimeout <= 0) ? 99999 : registrationTimeout;
                 player.addPotionEffect(new PotionEffect(PotionEffectType.BLINDNESS, blindTimeOut, 2));
             }
+
+            if (service.getProperty(RestrictionSettings.SPECTATE_STAND_LOGIN)) {
+                // The delay is necessary in order to make sure that the player is teleported to spawn and after authorization appears in the same place
+                bukkitService.runTaskLater(() -> {
+                    spectateLoginService.createStand(player);
+                }, 1);
+            }
+
             commandManager.runCommandsOnJoin(player);
         });
     }
--- a/src/main/java/fr/xephi/authme/service/SpectateLoginService.java
+++ b/src/main/java/fr/xephi/authme/service/SpectateLoginService.java
@@ -1,0 +1,86 @@
+package fr.xephi.authme.service;
+
+import fr.xephi.authme.settings.properties.RestrictionSettings;
+import org.bukkit.GameMode;
+import org.bukkit.Location;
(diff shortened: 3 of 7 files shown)


sgdc3:
Looks cool! I'll take a look asap, in the meantime @ljacqu @games647 @Hex3l what do you think about this concept+implementation?

src/main/java/fr/xephi/authme/process/join/AsynchronousJoin.java
```diff
@@ -191,6 +195,15 @@ private void processJoinSync(Player player, boolean isAuthAvailable) {
                 int blindTimeOut = (registrationTimeout <= 0) ? 99999 : registrationTimeout;
                 player.addPotionEffect(new PotionEffect(PotionEffectType.BLINDNESS, blindTimeOut, 2));
             }
+
+            if (service.getProperty(RestrictionSettings.SPECTATE_STAND_LOGIN)) {
+                // The delay is necessary in order to make sure that the player is teleported to spawn
```
sgdc3, line 200: What about listening at the player spawn/respawn event?

sgdc3 (changes requested):

src/main/java/fr/xephi/authme/settings/properties/RestrictionSettings.java
```diff
@@ -193,6 +193,22 @@ public final class RestrictionSettings implements SettingsHolder {
     public static final Property<Set<String>> UNRESTRICTED_INVENTORIES =
         newLowercaseStringSetProperty("settings.unrestrictions.UnrestrictedInventories");
 
+    @Comment({
+        "While in unregistered/logged out state, should players be set to spectator mode and",
+        "forced to spectate within a spawn point invisible armor stand, for 0 movement and head",
+        "pitch + yaw? may be more effective than 'allowMovement' at locking the player in place."
+    })
+    public static final Property<Boolean> SPECTATE_STAND_LOGIN =
+        newProperty("settings.restrictions.spectateStandLogin", true);
```
ljacqu, line 202: I think the default value should be `false` so that this doesn't change for everyone already using AuthMe

src/main/java/fr/xephi/authme/service/SpectateLoginService.java
```diff
@@ -0,0 +1,86 @@
...
+import org.bukkit.GameMode;
+import org.bukkit.Location;
+import org.bukkit.entity.ArmorStand;
+import org.bukkit.entity.Player;
+
+import javax.inject.Inject;
+import java.util.HashMap;
+import java.util.Map;
+
+public class SpectateLoginService {
```
ljacqu, line 13: Good name! Allows us to remember that this will set players to SPECTATOR game mode :)

src/main/java/fr/xephi/authme/service/SpectateLoginService.java
```diff
@@ -0,0 +1,86 @@
...
+            player.setSpectatorTarget(null);
+            player.setGameMode(gameModeMap.get(player));
+
+            gameModeMap.remove(player);
+            armorStands.remove(player);
+        }
+    }
+
+    public boolean hasStand(Player player) {
+        return armorStands.get(player) != null;
```
ljacqu, line 68: can be `armorStands.containsKey(player)`

src/main/java/fr/xephi/authme/settings/properties/RestrictionSettings.java
```diff
@@ -193,6 +193,22 @@ public final class RestrictionSettings implements SettingsHolder {
...
+
+    @Comment("Head Yaw position for 'spectateStandLogin'.")
+    public static final Property<Double> HEAD_YAW =
+        newProperty("settings.restrictions.headYaw", 0.0f);
+
+    @Comment("Head Pitch position for 'spectateStandLogin'.")
+    public static final Property<Double> HEAD_PITCH =
+        newProperty("settings.restrictions.headPitch", 0.0f);
```
ljacqu, line 210: Would it make sense to put these paths under a dedicated path? In the sense of `settings.restrictions.spectateStandLogin.pitch` (or .headPitch). With that naming, settings.restrictions.spectateStandLogin would have to become something like "settings.restrictions.spectateStandLogin.enabled"

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -508,4 +514,25 @@ public void onPlayerInventoryClick(InventoryClickEvent event) {
...
+        }
+    }
+
+    @EventHandler(ignoreCancelled = true, priority = EventPriority.HIGHEST)
+    public void onTeleport(PlayerTeleportEvent event) {
+        if (listenerService.shouldCancelEvent(event.getPlayer())
+            && event.getCause() == PlayerTeleportEvent.TeleportCause.SPECTATE
+            && event.getPlayer().getGameMode() == GameMode.SPECTATOR
+            && (settings.getProperty(RestrictionSettings.SPECTATE_STAND_LOGIN)
+            || spectateLoginService.hasStand(event.getPlayer()))) {
```
ljacqu, line 533: I guess this checks out since `listenerService.shouldCancelEvent(event.getPlayer())` will be false if the player is logged in, right? In other words, even when we enable a feature, players can still teleport to other players in spectate mode once they've logged in

src/main/java/fr/xephi/authme/service/SpectateLoginService.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private ArmorStand spawnStand(Location loc) {
+        double pitch = service.getProperty(RestrictionSettings.HEAD_PITCH);
+        double yaw = service.getProperty(RestrictionSettings.HEAD_YAW);
+        Location location = new Location(loc.getWorld(), loc.getX(), loc.getY(), loc.getBlockZ(),
+            (float) yaw, (float) pitch);
+
+        ArmorStand stand = location.getWorld().spawn(location, ArmorStand.class);
```
ljacqu, line 77: Does this cause issues if two players have the same location?

src/main/java/fr/xephi/authme/process/join/AsynchronousJoin.java
ljacqu: \+ doesn't the armor stand have to be deleted if the player disconnects as well?

src/main/java/fr/xephi/authme/service/SpectateLoginService.java
```diff
@@ -0,0 +1,86 @@
...
+     * Creates a stand for the player
+     *
+     * @param player the player
+     */
+    public void createStand(Player player) {
+        Location location = player.getLocation();
+        ArmorStand stand = spawnStand(location);
+
+        armorStands.put(player, stand);
+        gameModeMap.put(player, player.getGameMode());
```
ljacqu, line 31: What if the server shuts down while players are not logged in? I guess they'll be saved with the spectator game mode, and next time when they'll join, they'll already have spectator game mode and won't be able to leave it. I think this is quite relevant because it's not only an inconvenience to a player, it's giving them access to a game mode that many servers don't want to give to players... 
I could imagine trying to make use of the limbo player stuff or to maybe have an additional setting that will set players to survival instead of spectator when the code would want to. It's ugly but we've had to add a few of those failsafe properties in the past and people seem happy with it

ljacqu (commented):
Code looks great and you've found a great way to incorporate the changes into the existing codebase! Some unit tests would be nice :)

src/main/java/fr/xephi/authme/settings/properties/RestrictionSettings.java
mrchefran: The default value is now `false`

src/main/java/fr/xephi/authme/service/SpectateLoginService.java
mrchefran: Added a description of the class behavior

src/main/java/fr/xephi/authme/service/SpectateLoginService.java
mrchefran: Changed to it

src/main/java/fr/xephi/authme/listener/PlayerListener.java
mrchefran: These checks are designed to prevent unauthorized players from teleporting (using hotbar) while they are in spectator gamemode

src/main/java/fr/xephi/authme/service/SpectateLoginService.java
mrchefran: Apparently, when the plugin disabled, I can call a function that will remove all armorstands and return the players to their previous state. Is it possible to implement it this way?

src/main/java/fr/xephi/authme/process/join/AsynchronousJoin.java
mrchefran: > What about listening at the player spawn/respawn event?

Looks like this will also require a delay, because at this stage it's impossible to change the player gamemode

mrchefran: > + doesn't the armor stand have to be deleted if the player disconnects as well?

Armorstands are removed if an unauthorized player leaves the server

src/main/java/fr/xephi/authme/settings/properties/RestrictionSettings.java
mrchefran: Yes, it would. It looks much better this way. I put the properties in a separate path

src/main/java/fr/xephi/authme/service/SpectateLoginService.java
mrchefran: This does not affect the position of the stand in any way. The only thing is that players can see the stands of others if their view is directed low enough. The players don't see each other

src/main/java/fr/xephi/authme/service/SpectateLoginService.java
mrchefran: Could you please tell me if I can call SpectateLoginService in onDisable directly to remove armorstands, or is this not a good practice?

src/main/java/fr/xephi/authme/service/SpectateLoginService.java
ljacqu: So the first armor stand doesn't get replaced by the second one when the same location is used?

src/main/java/fr/xephi/authme/service/SpectateLoginService.java
ljacqu: Sorry for the late reply.
I don't see any better way. I think you could get the SpecateLoginService with `injector.getIfAvailable(SpectateLoginService.class)` in AuthMe#onDisable. But still, when a server unexpectedly shuts down (if it crashes or its process is terminated abruptly) we have the problem that players are potentially in spectator mode, and there are armor stands that no one tracks anymore. I don't know if the armor stands are an issue (I have zero experience with the way you spawn them). 
Wondering if we should have a property that specifies what game mode a player should change to if after authentication, we would put him into spectate game mode and/or to add the game mode to the LimboPlayer. I can elaborate on the whole limbo player thing if it helps

mrchefran: Untraceable armorstand is not as critical as players in Spectator gamemode. It would be really cool if it could be done with LimboPlayer

src/main/java/fr/xephi/authme/service/SpectateLoginService.java
mrchefran: No, it won't be replaced

src/main/java/fr/xephi/authme/service/SpectateLoginService.java
sgdc3: What about adding a "limboArmorStandUUID" field to LimboPlayer?

mrchefran: Do you mean adding a way to keep track of whether a player is assigned an armorstand?

sgdc3: @mrchefran yes, exactly

PR #2562, opened 2022-06-14
Add test for AuthMeApi.getLastLoginMillis

lacinoire:
Hey 😊
I want to contribute the following test:

Test that `result.getLastLoginTime("notAPlayer")` is null when `getLastLoginTime` is called with the parameter `playerName = "notAPlayer"`.
This tests the method [`AuthMeApi.getLastLoginMillis`](https://github.com/AuthMe/AuthMeReloaded/blob/3892bb692354d4591aac023f5eb05f68466c8353/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java#L202).
This test is based on the test [`shouldReturnInstanceOrNull`](https://github.com/AuthMe/AuthMeReloaded/blob/3892bb692354d4591aac023f5eb05f68466c8353/src/test/java/fr/xephi/authme/api/v3/AuthMeApiTest.java#L75).

Curious to hear what you think!

Also, is the description I provided of the test is helping you to answer to this pull request? Why (not)?

(I wrote this test as part of a research study at TU Delft. [Find out more](https://github.com/lacinoire/lacinoire/blob/main/README.md))

### DIFF
--- a/src/test/java/fr/xephi/authme/api/v3/AuthMeApiTest.java
+++ b/src/test/java/fr/xephi/authme/api/v3/AuthMeApiTest.java
@@ -242,6 +242,12 @@ public class AuthMeApiTest {
     }
 
     @Test
+    public void testGetLastLoginMillis() {
+        AuthMeApi result = AuthMeApi.getInstance();
+        assertThat(result.getLastLoginTime("notAPlayer"), nullValue());
+    }
+
+    @Test
     public void shouldHandleNullLastLoginTime() {
         // given
         String name = "John";


games647:
Good catch, thanks. This seems to be a nice tool. 

I'm curious how it generated the parameter input, if it did. If we **really** wanted to be precise, then `notAPlayer` is a valid player name although not existing. So `unknown`/`unknownPlayer` or even random invalid player name like `???` would fit. However this is only my opinion.

games647 (approved):

lacinoire:
Thanks @games647!

The name was added by myself. The tool originally generated `aw4Z#8{}YbgCZ5GeAZBX` and I changed it to something that I expected to be more readable. Good pointer to use a really invalid name 👍🏼

PR #2606, opened 2022-08-20
Add Checkstyle rule to disallow toLowerCase() and toUpperCase() without Locale

ljacqu:
Not ideal but better than nothing :)


sgdc3:
👍🏻

PR #2631, opened 2022-10-15
Update log4j

skbeh:


ljacqu:
Fairly sure they have to stay in sync with what Minecraft is using

skbeh:
@ljacqu I don't think so. The log4j depends are usually packed with the plugin so it's independent of Minecraft's log4j.

sgdc3:
Why? I see no reason to update the dependency as we use it as a "provided" resource.

PR #2642, opened 2022-11-19
Add haveibeenpwned password check

AlexProgrammerDE:
This pull request adds a haveibeenpwned.com password check before allowing a user to register with a password.
This will hopefully force users to not use weak passwords, which are easy to crack due to comparing the stored hash with password lists. The password is sent hashed and only the first five characters of the hash to haveibeenpwned. Then the response is validated by AuthMe on the server, which is going through ~500 hashes returned by the API. So this is a very secure way of checking for how secure a password is.
example message with weak password

References:
- [Example API call](https://api.pwnedpasswords.com/range/F46FA)
- [Main password check website](https://haveibeenpwned.com/Passwords)
- [API Documentation](https://haveibeenpwned.com/API/v3#PwnedPasswords)

It appears that there is a paid API, but from what I've seen, it is only for account breaches where you search by E-Mail, not by password. So I don't think there will be any rate limits this hits.
My discord is Pistonmaster#0001 (In AuthMe support discord server), let me know if there should be something changed here.

### DIFF
--- a/src/main/java/fr/xephi/authme/message/MessageKey.java
+++ b/src/main/java/fr/xephi/authme/message/MessageKey.java
@@ -80,7 +80,7 @@ public enum MessageKey {
     /** The chosen password isn't safe, please choose another one... */
     PASSWORD_UNSAFE_ERROR("password.unsafe_password"),
 
-    /** Your chosen password is not secure. It was used %pwned_count times already! Please use a stronger password... */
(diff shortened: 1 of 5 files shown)


AlexProgrammerDE:
@Xephi hello, what is the reason for you closing my pull request? I still want this feature shipped.

AlexProgrammerDE:
@Xephi I can redo this PR if the code wasn't to your liking.

PR #2660, opened 2022-12-23
Issues/2659: Fix password comparison.

### LINKED ISSUE #2659: Conversion with CrazyLogin
### What behaviour is observed?

I have a database from CrazyLogin in flat format. I copied it to "plugins/AuthMe", configured "Converter.CrazyLogin.fileName" in "config.yml" according to the name of the database file and set "settings.security.PasswordHash" to "CRAZYCRYPT1". Next, I run the server version "1.18.2" with the plugin version "5.6.0-beta2" and enter the command "authme converter crazylogin". The server responds:
```
[23:08:26] [Server thread/INFO]: [AuthMe] Successfully started crazylogin
[23:08:26] [Craft Scheduler Thread - 1/INFO]: [AuthMe] CrazyLogin database has been imported correctly
```
I log on to the server under the username "Freeman" and the server asks me to log in, that is, my username has really been converted and I do not need to register. I enter the password from the CrazyLogin database "TEST_TEST", but the server writes that my password is incorrect.

I tried debugging the code and saw that the hash of my password really didn't match. The password hash in CrazyLogin depends on the user's login, but for some reason the plugin translates my login to lowercase, that is, it turns out "freeman", and it also turns out another password hash.

### Expected behaviour

I was registered in CrazyLogin under the username "Freeman" with the password "TEST_TEST". After converting the database, I expect that AuthMe will consider my login-password bundle to be correct.

### To Reproduce

In the CrazyLogin database there is a user with the username "Freeman" and the password "TEST_TEST". He successfully passes authorization with the CrazyLogin plugin.

1. Moved the CrazyLogin databases to "plugins/AuthMe";
2. In "config.yml" set for the parameter "settings.security.PasswordHash" value "CRAZYCRYPT1";
3. Started the server;
4. Entered the command "authme converter crazylogin";
5. I log on to the server with the username "Freeman";
6. I enter "/login TEST_TEST";
7. I am disconnected from the server with the message "Wrong password!".

### Plugin list

Plugins (1): AuthMe

### Server Implementation

Standalone server (no proxy)

### Database Implementation

SQLite

### AuthMe Version

Version: AuthMeReloaded v5.6.0-beta2 (build: 2453)

### Error log (if applicable)

_No response_

### Configuration

https://gist.github.com/Freemandns/e4e562725905ec8b16f6c2c4619eed01

Freemandns:
Hello. This fixes #2659 . I don't know Java well, so maybe I did something wrong.

### DIFF
--- a/src/main/java/fr/xephi/authme/security/PasswordSecurity.java
+++ b/src/main/java/fr/xephi/authme/security/PasswordSecurity.java
@@ -82,9 +82,8 @@ public class PasswordSecurity implements Reloadable {
      * @return True if the password matches, false otherwise
      */
     public boolean comparePassword(String password, HashedPassword hashedPassword, String playerName) {
-        String playerLowerCase = playerName.toLowerCase(Locale.ROOT);
-        return methodMatches(encryptionMethod, password, hashedPassword, playerLowerCase)
-            || compareWithLegacyHashes(password, hashedPassword, playerLowerCase);
+        return methodMatches(encryptionMethod, password, hashedPassword, playerName)
+            || compareWithLegacyHashes(password, hashedPassword, playerName);
     }
 
     /**
--- a/src/test/java/fr/xephi/authme/security/PasswordSecurityTest.java
+++ b/src/test/java/fr/xephi/authme/security/PasswordSecurityTest.java
@@ -34,7 +34,7 @@ import static com.google.common.collect.Sets.newHashSet;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.equalToIgnoringCase;
 import static org.hamcrest.Matchers.instanceOf;
-import static org.junit.Assert.assertThat;
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.BDDMockito.given;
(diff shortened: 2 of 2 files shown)


sgdc3:
Please explain how this fixes the issue, it isn't clear to me.

Freemandns:
I have already described problem #2659 . I used the CrazyLogin plugin, which hashes passwords using the CRAZYCRYPT1 algorithm, which in turn uses the user's nickname as a salt. Because of this, the same password will have a different hash if the user uses his nickname in different registers.

For example, my nickname is Freeman, and my password is TEST_TEST.
**CrazyLogin**: hashes my password using the CRAZYCRYPT1 algorithm using the "Freeman" salt, it turns out - _763304530c0eed64841526d157ec1f9b7fc20f7f73da39fbf66ebd6bc228b5b72a945ae2048fb11bd74a772f79beb10baaf1009f9b1222be667d1474b41fbf45_
**AuthMeReloaded** (configured for hashing using the CRAZYCRYPT1 algorithm): converts my nickname to a low case, it turns out "freeman", hashes my password using the CRAZYCRYPT1 algorithm using the "freeman" salt, it turns out - _86ec6077c9a4db611481863f87005f0caf8c7e663ea51e0f88adf2b6929dc1729055c798c6fa4664939c9e342b7e065217c7921e8a9625a7ca7cd0f8f090466c_

Since the hashes do not match, after converting the database from CrazyLogin to AuthMeReloaded, my password is always incorrect.

I removed the conversion to lowercase and now this fix works well on my server.

ljacqu:
Thanks for the background. But this change breaks all other EncryptionMethod classes that rely on the passed in username being lowercase

PR #2720, opened 2023-06-25
Config option to disable GeoIP database

3meraldK:


ljacqu:
I don't think the constructor is the correct place to check for this, but rather in `isDataAvailable()`. It's not just called by the constructor but also in lookup methods, so it would be better to properly catch it in `isDataAvailable` and to avoid conditional logic in the constructor.

sgdc3 (approved):

3meraldK:
Requesting @ljacqu to review

ljacqu:
Two unit tests fail, and I think we don't need the `enabled` field since all public data methods go through `getCountry()` -> I will fix this

ljacqu:
The changes are now on master. Thanks!

PR #2735, opened 2023-07-30
Actually finish the Lithuanian translation

Obydux:
The current Lithuanian translation wasn't actually finished and it didn't make use of letters found in the Lithuanian alphabet.


ljacqu:
Wonderful work—thank you!

ljacqu (approved):

PR #2745, opened 2023-08-31
SpectateLogin implementation

Bagietas:
Re-opened https://github.com/AuthMe/AuthMeReloaded/pull/2552 cause I don't have access to modify any of that.

### DIFF
--- a/src/main/java/fr/xephi/authme/AuthMe.java
+++ b/src/main/java/fr/xephi/authme/AuthMe.java
@@ -24,6 +24,7 @@ import fr.xephi.authme.security.crypts.Sha256;
 import fr.xephi.authme.service.BackupService;
 import fr.xephi.authme.service.BukkitService;
 import fr.xephi.authme.service.MigrationService;
+import fr.xephi.authme.service.SpectateLoginService;
 import fr.xephi.authme.service.bungeecord.BungeeReceiver;
 import fr.xephi.authme.service.yaml.YamlParseException;
 import fr.xephi.authme.settings.Settings;
@@ -69,6 +70,7 @@ public class AuthMe extends JavaPlugin {
     private Injector injector;
     private BackupService backupService;
     private ConsoleLogger logger;
+    private SpectateLoginService spectateLoginService;
 
     /**
      * Constructor.
@@ -246,6 +248,7 @@ public class AuthMe extends JavaPlugin {
         bukkitService = injector.getSingleton(BukkitService.class);
         commandHandler = injector.getSingleton(CommandHandler.class);
         backupService = injector.getSingleton(BackupService.class);
+        spectateLoginService = injector.getSingleton(SpectateLoginService.class);
 
         // Trigger instantiation (class not used elsewhere)
         injector.getSingleton(BungeeReceiver.class);
@@ -315,6 +318,8 @@ public class AuthMe extends JavaPlugin {
 
         // Wait for tasks and close data source
         new TaskCloser(this, database).run();
+
+        spectateLoginService.removeArmorstands();
 
         // Disabled correctly
         Consumer<String> infoLogMethod = logger == null ? getLogger()::info : logger::info;
--- a/src/main/java/fr/xephi/authme/listener/PlayerListener.java
+++ b/src/main/java/fr/xephi/authme/listener/PlayerListener.java
@@ -11,6 +11,7 @@ import fr.xephi.authme.process.Management;
 import fr.xephi.authme.service.AntiBotService;
 import fr.xephi.authme.service.BukkitService;
 import fr.xephi.authme.service.JoinMessageService;
+import fr.xephi.authme.service.SpectateLoginService;
 import fr.xephi.authme.service.TeleportationService;
 import fr.xephi.authme.service.ValidationService;
 import fr.xephi.authme.settings.Settings;
@@ -19,6 +20,7 @@ import fr.xephi.authme.settings.properties.HooksSettings;
 import fr.xephi.authme.settings.properties.RegistrationSettings;
 import fr.xephi.authme.settings.properties.RestrictionSettings;
 import org.bukkit.ChatColor;
+import org.bukkit.GameMode;
 import org.bukkit.Location;
 import org.bukkit.entity.HumanEntity;
 import org.bukkit.entity.Player;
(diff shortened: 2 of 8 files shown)


ljacqu:
- Unit tests are missing
- If the server crashes and players are not authenticated, players will be stuck in spectator gamemode (see comment thread in the original PR referencing "limbo player")

src/main/java/fr/xephi/authme/settings/properties/RestrictionSettings.java
```diff
@@ -193,6 +193,22 @@ public final class RestrictionSettings implements SettingsHolder {
...
+    })
+    public static final Property<Boolean> SPECTATE_STAND_LOGIN =
+        newProperty("settings.restrictions.spectateStandLogin.enabled", false);
+
+    @Comment("Head Yaw position for 'spectateStandLogin'.")
+    public static final Property<Double> HEAD_YAW =
+        newProperty("settings.restrictions.spectateStandLogin.headYaw", 0.0f);
+
+    @Comment("Head Pitch position for 'spectateStandLogin'.")
+    public static final Property<Double> HEAD_PITCH =
```
ljacqu, line 209: I think this should work with the current version of ConfigMe (ignore this comment if it doesn't)

```suggestion
    public static final Property<Float> HEAD_PITCH =
```

Also the naming is a bit weird, I don't know that `RestrictionSettings#HEAD_PITCH` refers to the spectate stand feature, and as a user I'm not sure I'll understand what this setting represents. Is it possible to elaborate in the comment? I would make a suggestion but I'm actually not really sure what the yaw & pitch settings are for

src/main/java/fr/xephi/authme/AuthMe.java
```diff
@@ -316,6 +319,8 @@ public void onDisable() {
         // Wait for tasks and close data source
         new TaskCloser(this, database).run();
 
+        spectateLoginService.removeArmorstands();
```
ljacqu, line 322: Potential null pointer if the plugin crashes before this service is set on L251

src/main/java/fr/xephi/authme/listener/PlayerListener.java
```diff
@@ -376,6 +382,34 @@ public void onPlayerRespawn(PlayerRespawnEvent event) {
         if (spawn != null && spawn.getWorld() != null) {
             event.setRespawnLocation(spawn);
         }
+
+        if (settings.getProperty(RestrictionSettings.SPECTATE_STAND_LOGIN)
+            || spectateLoginService.hasStand(event.getPlayer())) {
+            // If a player is dead, no stand will be created for him
+            // therefore we can be sure that there will not be two stands
+            bukkitService.runTaskLater(() -> spectateLoginService.createStand(event.getPlayer()), 1L);
```
ljacqu, line 390: I can't confirm the comment about stands not being created if a player is dead but maybe I'm missing the relevant point, but interestingly the logic here says if the setting is enabled, _or_ the player has a stand, then a stand is created? I don't think that logic adds up

src/main/java/fr/xephi/authme/settings/properties/RestrictionSettings.java
Bagietas: I don't know If i understood you correctly, but `HEAD_PITCH` and `HEAD_YAW` are under `spectateStandLogin` so It's obvious what are these settings for.

ljacqu: Thanks for changing the comment. I also meant the field name, `RestrictionSettings.HEAD_PITCH` in the code sounds very generic. Could we maybe name it `SPECTATE_HEAD_PITCH`, or maybe you have a better idea?

src/main/java/fr/xephi/authme/AuthMe.java
ljacqu: 1. The field on the class is no longer needed (an IDE shows an unused warning)
2. `injector.createIfHasDependencies(SpectateLoginService.class)` might return null, so the issue of a potential null pointer still remains. And I think it would be better to use `injector#getIfAvailable`

src/main/java/fr/xephi/authme/listener/PlayerListener.java
ljacqu: Now an armor stand is created for a player every time he respawns, regardless of whether an armor stand already exists or not

ljacqu (changes requested):
Your code doesn't compile btw

Bagietas:
> > > Unit tests are missing
> > 
> > 
> > What do you mean by this? (Never used tests)
> 
> Every class has a "Test" class associated to it, e.g. BukkitService has BukkitServiceTest. These are _unit tests_, which ensure that refactorings and similar don't break things in unexpected ways. Searching for JUnit should also provide additional information.
> 
It's my first time I do something like tests. That's my try to implement it but I don't know if it's correct.

> > > If the server crashes and players are not authenticated, players will be stuck in spectator gamemode (see comment thread in the original PR referencing "limbo player")
> > 
> > 
> > Implemented it.
> 
> I think it would be important to test different scenarios, like if existing limbo player data exists before these changes, there might be limbo player data already stored on disk. Since no "gamemode" exists, it's null, which I think will cause further issues down the line when we try to set the gamemode to `null` on a player. I don't fully remember the whole lifecycle of the LimboPlayer (and I'm going abroad soon, so a little pressed for time this week) but I think this would be nice to really test in detail to make sure this won't break players' data. It might make sense to only touch the gamemode if the player is currently in spectator mode, the spectating setting is enabled, and we have another gamemode on the limbo player. Or maybe to have a config like we have for many of the other limbo stuff to allow it to be enabled and disabled. Someone else from the AuthMe team might have some good inputs here.

If I understood you correctly I've added what you've meant.

src/main/java/fr/xephi/authme/listener/PlayerListener.java
Bagietas: Done.

src/main/java/fr/xephi/authme/AuthMe.java
Bagietas: Done.

src/main/java/fr/xephi/authme/settings/properties/RestrictionSettings.java
Bagietas: Renamed.

PR #2778, opened 2024-02-21
Update PostgreSQL to resolve CVE-2024-1597

HaHaWTH:
This pull request update postgresql to 42.7.2


games647:
This only affects projects using the non-default `PreferQueryMode=SIMPLE` option. We don't use it in our code and AFAIK no dependency sets this option either. However, it could make sense to update it anyway to silence the warning.

PR #2823, opened 2024-05-20
Fix: Teleporting Players to Spawn on Login with teleportUnAuthedToSpawn=true

Subu19:
Issue: When changing the config option teleportUnAuthedToSpawn=true during midgame, players were teleported to spawn on their next login. It only happens once after the update.

Patch: AuthMe now checks the player's last logout location from Essentials if it is not found in the AuthMe database. This ensures players are teleported to their last known location instead of the spawn point when the configuration is changed mid-game.


F3F5:
if there are no essentials

Subu19:
> if there are no essentials

You are right, this solution only works if Essentials is installed.
The original code looks okay, it does try to teleport to LimboPlayer's original location after logging in if the last location is not found in the authme's database. I don't know if this was intentional or not, LamboPlayer is initiated after teleporting the player. We could fix this issue if we store the original location of the player in LimboPlayer.

F3F5:
https://github.com/F3F5/SpawnAuth my solution

PR #2962, opened 2026-04-20
AuthMe 6.0.0

Xephi:
## `f93800cc4` `refactor: AuthMe is now multi-platform`

- Converts the repository from a single plugin layout to a multi-module Maven build.
- Moves the existing implementation into `authme-core` and introduces the first platform deliverables (`authme-spigot-legacy`, `authme-spigot-1.21`, `authme-paper`).
- Establishes the initial `PlatformAdapter` split and folds the version bump into the structural migration.

`PlatformAdapter` is the seam between the platform-agnostic logic in `authme-core` and the server-specific implementation shipped by each deliverable module. At startup, AuthMe loads exactly one adapter from the plugin jar, registers it in the DI container both as `PlatformAdapter` and as its focused sub-interfaces, then lets core services call those narrow contracts (`TeleportAdapter`, `ChatAdapter`, `EventRegistrationAdapter`, `SchedulingAdapter`, etc.) without branching on Paper vs Spigot in business code.

In practice, an adapter implementation answers these questions:

1. **Is the current server compatible?** via `getCompatibilityError()`
2. **Which platform-specific listeners or overrides should be exposed?** via methods such as `getAdditionalListeners()`

Example of a concrete adapter implementation:

```java
public class PaperPlatformAdapter extends AbstractPaperPlatformAdapter {

    @Override
    public String getPlatformName() {
        return "paper-1.21";
    }

    @Override
    public String getCompatibilityError() {
        return getCompatibilityError(
            "This AuthMe Paper build requires the Paper 1.21.11+ API.",
            "io.papermc.paper.event.player.AsyncChatEvent",
            "io.papermc.paper.event.player.AsyncPlayerSpawnLocationEvent",
            "io.papermc.paper.event.player.PlayerOpenSignEvent");
    }

    @Override
    public List<Class<? extends Listener>> getAdditionalListeners() {
        return Arrays.asList(
            PaperChatListener.class,
            PaperPlayerSpawnLocationListener.class,
            PlayerOpenSignListener.class);
    }
}
```

## `2a6c9f612` `feat: Dialog implementation (Spigot/PaperMC 1.21.8+)`

- Adds `DialogAdapter` to the platform abstraction.
- Implements login/register dialogs for Spigot and Paper through dedicated helper classes.
- Wires dialog support into join and registration flows.

## `6ff78b725` `test: Update to Junit/Mockito 5`

- Migrates the test suite to JUnit Jupiter and Mockito Jupiter.
- Introduces the local Jupiter helpers used by the suite (`DelayedInjectionExtension`, `TempFolder`).
- Updates module test dependencies and listener/platform tests to the new stack.

## `0bdfaf9c6` `refactor: Move loading safeguards to each platform implementations`

- Pushes compatibility checks into the platform adapters instead of keeping them in core bootstrap code.
- Adds adapter-specific startup validation and the Paper spawn-location listener needed by that flow.
- Keeps startup failure reasons close to the platform that owns them.

## `e28ded77f` `feat: Register commands through Brigadier for Paper`

- Adds `CommandRegistrationAdapter` to the platform abstraction.
- Introduces Paper-side Brigadier registration support and associated tests.
- Keeps core command handling authoritative while letting Paper expose richer command metadata.

## `7be399af3` `fix: Asynchronous mail verification`

- Fixes verification-code handling so asynchronous mail verification no longer races or loses state.
- Tightens the corresponding tests around the verification manager behavior.

## `bb5f6f1c2` `refactor: rethink spawn/respawn/join location logic`

- Reworks teleport and spawn selection logic across join, respawn, quit, and limbo flows.
- Extends the platform abstraction where teleport/spawn behavior diverges.
- Updates tests around `TeleportationService`, `SpawnLoader`, and limbo persistence expectations.

## `5977687e1` `refactor: Abstract scheduling for cross-platform thread management`

- Introduces `SchedulingAdapter` and `CancellableTask` in `authme-core`.
- Moves entity/global thread ownership decisions behind platform-specific scheduling APIs.
- Adjusts core services and async/sync process hand-offs so thread-sensitive work is dispatched correctly.

## `5dfb395f3` `refactor: Extract shared Paper code into authme-paper-common`

- Extracts Paper/Folia shared listeners, dialog helpers, and command-registration helpers into a common module.
- Makes `authme-paper` depend on that shared module instead of owning duplicated implementation details.
- Leaves Paper-specific entrypoints thin and focused on bootstrap concerns.

## `3ae4257ae` `feat: Add Folia platform support`

- Adds the `authme-folia` deliverable module, metadata, and ServiceLoader registration.
- Implements the Folia platform adapter and listener layer on top of the new scheduling abstraction.
- Covers the new module with dedicated platform/listener tests.

## `f994d9e5a` `docs: Update documentation tooling/templates and re-generate`

- Refreshes the documentation templates and tooling under `authme-core/src/test/java/tools/docs/`.
- Re-generates the published files under `docs/` to match the new platform/module architecture.

## `7c55478b4` `chore: Build java 21 modules only when java 21 is available`

- Gates the Java 21 deliverable modules behind a Java-version activation block in the parent build.
- Keeps the Java 17-compatible modules buildable and testable without requiring the full Java 21 reactor.
 
## `3931107b8` `fix(platform): migrate Paper/Folia login handling to split events`

- Moves the Paper/Folia login-time checks away from the deprecated `PlayerLoginEvent`.
- Adds a shared `PaperLoginValidationListener` that uses `PlayerConnectionValidateLoginEvent` for single-session checks and `PlayerServerFullCheckEvent` for full-server/VIP handling.
- Keeps Spigot on the legacy `PlayerLoginEvent` path while disabling that path on Paper-derived platforms.

## `2c6c384b1` `feat: Add PBKDF2Base64 hash algorithm and Auth+ converter`

- Introduces `AbstractPbkdf2` as a shared base for PBKDF2-based encryption methods, factoring out `numberOfRounds` configuration and the `PBKDF2Engine` derivation/verification helpers.
- Refactors `Pbkdf2` to extend `AbstractPbkdf2` instead of `HexSaltedMethod` directly.
- Adds `Pbkdf2Base64`: PBKDF2-HmacSHA256 with Base64-encoded salt and hash, format `pbkdf2$<iter>$<salt_b64>$<hash_b64>`, registered as `PBKDF2BASE64` in `HashAlgorithm`.
- Adds `AuthPlusConverter` to migrate accounts from the Auth+ plugin (`plugins/Auth/players.yml`). Resolves UUID to player name via the Bukkit offline-player cache. Requires `passwordHash: PBKDF2BASE64` and `pbkdf2Rounds: 120000` in AuthMe's config before running.
- Adds `docs/converters.md`, a hand-maintained reference page documenting all available converters and their prerequisites.

## `8c8b025a3` `feat(config): split timeout into loginTimeout and registerTimeout`

- Replaces the single `settings.restrictions.timeout` property with two independent properties: `settings.restrictions.loginTimeout` and `settings.restrictions.registerTimeout`.
- Both default to `30` seconds (the previous default), so behaviour is unchanged out of the box.
- Adds an automatic migration in `SettingsMigrationService` that reads the old `timeout` value and writes it to whichever of the two new keys are missing, preserving users' customised values across upgrade.
- Updates `LimboPlayerTaskManager`, `LimboService`, `AsynchronousJoin`, and the logout/unregister processes to inject the two distinct settings.
- Marks `settings.restrictions.timeout` as an obsolete key that gets cleaned up on the next config rewrite.

## `96a2ff801` `feat(i18n): Add per-player locale and fallback on server global messages on need`

- Adds `PlayerLocaleResolver`, which maps Minecraft client locale strings (e.g. `fr_fr`, `zh_cn`) to AuthMe language codes (`fr`, `zhcn`), covering special cases such as `pt_br` → `br` and the Chinese regional variants.
- Extends `AbstractMessageFileHandler` with a per-language `FileConfiguration` cache: language files are copied from the JAR on first access and kept in memory, and the cache is cleared on `/authme reload`. `getMessage(key, language)` falls back to the server-configured language when the requested locale is unavailable or missing a key.
- Updates `Messages` to resolve each player's client locale through `PlayerLocaleResolver` and route the lookup through the new per-language cache, while non-player senders continue to use the global language.
- Adds `settings.perPlayerLocale` (default `true`) to `PluginSettings`, allowing servers with fully customised single-language message files to opt out and send all players the same language.


Xephi:
We're now producing jar for different platforms : 

- AuthMe-6.0.0-SNAPSHOT-Spigot-Legacy.jar
- AuthMe-6.0.0-SNAPSHOT-Spigot-1.21.jar
- AuthMe-6.0.0-SNAPSHOT-Paper.jar
- AuthMe-6.0.0-SNAPSHOT-Folia.jar

lokspel:
Wow.

lokspel:
Why the listeners isnt seperate as well

Xephi:
> Why the listeners isnt seperate as well

What do you mean ? @lokspel 
We have a default implementation for Listeners. We extract only Listeners that need to be override by other server implementations.

lokspel:
Oh, okay. Hopefully it will be merged to the main branch anytime soon, its good update.

Xephi:
> Oh, okay. Hopefully it will be merged to the main branch anytime soon, its good update.

I am awaiting the review of the amazing people who maintained the project instead of me over the past few years.

authme-core/src/main/java/fr/xephi/authme/settings/properties/RegistrationSettings.java
```diff
@@ -107,6 +107,13 @@ public final class RegistrationSettings implements SettingsHolder {
     public static final Property<Boolean> PREVENT_OTHER_CASE =
         newProperty("settings.preventOtherCase", true);
 
+    @Comment({
+        "Show a graphical dialog UI for login/register instead of chat messages.",
+        "Requires Minecraft 1.21.6+ on Spigot, or 1.21.11+ on Paper.",
```
sgdc3, line 112: Discrepancy with the commit message (feat: Dialog implementation (Spigot/PaperMC 1.21.8+)

authme-core/src/main/java/fr/xephi/authme/AuthMe.java
```diff
@@ -289,8 +289,6 @@ void registerEventListeners(Injector injector) {
         // All supported versions are >= 1.16, so these listeners are always safe to register
         pluginManager.registerEvents(injector.getSingleton(PlayerListener19.class), this);
```
sgdc3: We should merge them into the main player listener, as we only support MC 1.16+

authme-tools/src/test/java/tools/filegeneration/GeneratePluginYml.java
sgdc3, line 1: All the tools should be moved/updated accordingly, this one for example was used to generate the plugin.yml file

sgdc3 (changes requested):

authme-core/src/main/java/fr/xephi/authme/settings/properties/RegistrationSettings.java
Xephi: Yeah, but the config message is right. just they consolidate the api in 1.21.8+, so as a developer, you prefer to use the latest ver.

authme-core/src/main/java/fr/xephi/authme/AuthMe.java
Xephi: Fixed by 75ed08e9c5a391991b3447cf164d1bb68260b2fc

authme-tools/src/test/java/tools/filegeneration/GeneratePluginYml.java
Xephi: Hate you x')
https://github.com/AuthMe/AuthMeReloaded/pull/2962/changes/b3e794b0ea8c999bbbb13fdd271d2030d6857a77

PR #2982, opened 2026-04-22
feat!: replace ProtocolLib integration with PacketEvents

Xephi:
PacketEvents becomes the single packet interception backend so inventory
 protection and tab-complete blocking follow the new platform adapter
 lifecycle across all deliverables.

 BREAKING CHANGE: features that previously depended on ProtocolLib now
 require PacketEvents, and plugin manifests, docs, and test fixtures have
 been updated to reflect the new optional dependency


lokspel:
When it can be merged to the main

Xephi:
> When it can be merged to the main

When it would be tested :D

PR #3019, opened 2026-05-02
feat(premium): cryptographic session verification for premium bypass

Xephi:
## Requirements

- **PacketEvents** must be installed on the server, **unless** you
  are using proxy mode (see "Behind a proxy" on docs/premium.md).  

## How it works

When a premium-enrolled player connects, AuthMe intercepts the Minecraft login handshake at
the packet level:

```mermaid
sequenceDiagram
    autonumber
    participant C as Client
    participant S as Server (AuthMe + PacketEvents)
    participant M as Mojang

    C->>S: LOGIN_START (name)
    Note over S: ① DB: isPremium → true (async)
    S->>C: ENCRYPTION_REQUEST (RSA-1024 Key + Random Token)
    
    rect rgb(240, 240, 240)
        Note right of C: Client starts encrypting outbound traffic
        C->>M: POST /session/minecraft/join
        C->>S: ENCRYPTION_RESPONSE (enc sharedSecret + Token)
    end

    Note over S: ② RSA-decrypt sharedSecret (sync/event-loop)
    Note over S: ③ Install AES/CFB8 Netty ciphers (sync, client sends encrypted from here)
    
    S->>M: ④ Verify token + GET /hasJoined (async)
    M-->>S: {uuid, name, properties}
    
    Note over S: ⑤ Store verified UUID (60s TTL)
    Note over S: ⑥ Re-inject LOGIN_START

    S->>C: Login Success

    rect rgb(230, 245, 230)
        Note over S: AsynchronousJoin: 
        Note over S: getVerifiedUuid(name) == auth.getPremiumUuid() -> Success / Auto-Login
    end
```


authme-core/src/main/java/fr/xephi/authme/service/PremiumLoginVerifier.java
```diff
@@ -0,0 +1,199 @@
...
+    private static final long VERIFIED_TTL_MS = 60_000L;
+
+    private final ConsoleLogger logger = ConsoleLoggerFactory.get(PremiumLoginVerifier.class);
+
+    private final KeyPair rsaKeyPair;
+    private final SecureRandom secureRandom;
+    private final MojangApiService mojangApiService;
+
+    /** Keyed by connection address (ip:port). */
+    private final ConcurrentHashMap<String, PendingVerification> pending = new ConcurrentHashMap<>();
```
TuxCoding, line 47: Couldn't this cause memory leaks? For example if a malicious user sends only `LOGIN_START`, but terminates the connection so that no EncryptionResponse packet is sent.

authme-bungee/src/main/java/fr/xephi/authme/bungee/BungeeProxyBridge.java
```diff
@@ -158,6 +173,22 @@ public void onPluginMessage(PluginMessageEvent event) {
             logger.info("Auto-login ACK received for " + parsedMessage.playerName()
                 + " from server '" + server.getInfo().getName() + "'");
             cancelPendingLogin(parsedMessage.playerName());
+        } else if (PREMIUM_SET_MESSAGE.equals(parsedMessage.typeId())) {
+            premiumUsernames.add(parsedMessage.playerName());
+            logger.fine("Premium enabled for '" + parsedMessage.playerName() + "' (proxy cache updated)");
```
TuxCoding: Use parameter logging, so it reduces the overhead if that log level is disabled

```suggestion
            logger.fine("Premium enabled for {} (proxy cache updated)", parsedMessage.playerName());
```

README.md
```diff
@@ -58,6 +58,7 @@ You can also create your own translation file and, if you want, you can share it
   <li>Graphical login/register dialogs, with optional Paper/Folia pre-join dialogs</li>
   <li>Restricted users (associate a username with an IP)</li>
   <li>Protect player's inventory until correct authentication (requires PacketEvents)</li>
+  <li><strong>Premium bypass: Mojang-account holders skip password auth via cryptographic session verification (requires PacketEvents)</strong></li>
```
TuxCoding: Maybe this should be simplified for non-technical users ;) 

```suggestion
  <li><strong>Premium bypass: Mojang-account holders skip password auth (requires PacketEvents)</strong></li>
```

or 


```suggestion
  <li><strong>Premium bypass: Mojang-account holders skip password auth via emulating Minecraft authentication (requires PacketEvents)</strong></li>
```

authme-core/src/main/resources/messages/messages_en.yml
```diff
@@ -212,3 +212,22 @@ admin:
     first_set_fail: '&cSetFirstSpawn has failed, please retry.'
     not_defined: '&cSpawn has failed, please try to define the spawn.'
     first_not_defined: '&cFirst spawn has failed, please try to define the first spawn.'
+
+# Premium mode
+premium:
+  feature_disabled: '&cPremium mode is not enabled on this server.'
+  account_not_found: '&cNo premium Minecraft account found for your username.'
```
TuxCoding: There is no non-premium account, isn't there?

```suggestion
  account_not_found: '&cNo Minecraft account found for your username.'
```

or similar to the message below at `account_not_found`


```suggestion
  account_not_found: '&cNo Mojang account found for your username.'
```

authme-core/src/main/java/fr/xephi/authme/settings/properties/DatabaseSettings.java
```diff
@@ -138,6 +138,11 @@ public final class DatabaseSettings implements SettingsHolder {
     public static final Property<String> MYSQL_COL_PLAYER_UUID =
         newProperty( "DataSource.mySQLPlayerUUID", "" );
 
+    @Comment({"Column for storing the Mojang UUID of premium players",
+        "(null if premium mode is off for this player)"})
+    public static final Property<String> MYSQL_COL_PREMIUM_UUID =
+        newProperty("DataSource.mySQLColumnPremiumUUID", "premium_uuid");
```
TuxCoding: We used camelCase in other columns. So maybe here too?


```suggestion
        newProperty("DataSource.mySQLColumnPremiumUUID", "premiumUUID");
```

authme-velocity/src/main/java/fr/xephi/authme/velocity/VelocityProxyBridge.java
```diff
@@ -64,6 +72,35 @@ final class VelocityProxyBridge {
...
+    void onLogin(LoginEvent event) {
+        Player player = event.getPlayer();
+        if (!player.isOnlineMode()) {
+            return;
+        }
+        markProxyVerifiedPremium(normalizeName(player.getUsername()));
+    }
+
+    /**
+     * Fallback hook on {@link PostLoginEvent}: if the player has a UUID v4 (Mojang-issued), they
```
TuxCoding, line 101: What is the motivation for this fallback? Was there a workaround necessary? I think this could go wrong if another plugin injects a fake UUID. Then wouldn't we enable premium logins automatically? 

I mean it's unlikely, but plugins do all kind of things. If out of scope, then we can ignore it.

authme-paper-common/src/main/java/fr/xephi/authme/listener/PaperDialogFlowListener.java
```diff
@@ -294,6 +303,26 @@ private void completeRegisterResponse(UUID playerId, String kickMessage) {
...
+            return playerId.equals(auth.getPremiumUuid());
+        }
+        // UUID v3 (offline): check if PacketEvents has already completed verification.
+        UUID verifiedUuid = premiumLoginVerifier.getVerifiedUuid(playerName);
+        if (verifiedUuid != null) {
+            return verifiedUuid.equals(auth.getPremiumUuid());
+        }
+        // UUID is still offline at the configure phase but the player is enrolled as premium.
+        // Proxy UUID forwarding may not have been applied yet — skip the blocking pre-join dialog
+        // and let AsynchronousJoin.canBypassWithPremium() do the definitive check once the player
```
TuxCoding, line 323: Considering this comment section, the method doesn't check if the authentication can be **bypassed**, because the verification step is delayed to a different section. Therefore, we should maybe rename the method to something like `shouldSkipDialog`, because the current method name could be dangerous if relied on the premium state?

authme-velocity/src/main/java/fr/xephi/authme/velocity/VelocityProxyBridge.java
```diff
@@ -172,6 +210,22 @@ void onPluginMessage(PluginMessageEvent event) {
...
                 parsedMessage.playerName(), serverName);
             cancelPendingLogin(parsedMessage.playerName());
+        } else if (PREMIUM_SET_MESSAGE.equals(parsedMessage.typeId())) {
+            premiumUsernames.add(parsedMessage.playerName());
+            logger.debug("Premium enabled for '{}' (proxy cache updated)", parsedMessage.playerName());
+        } else if (PREMIUM_UNSET_MESSAGE.equals(parsedMessage.typeId())) {
+            premiumUsernames.remove(parsedMessage.playerName());
+            logger.debug("Premium disabled for '{}' (proxy cache updated)", parsedMessage.playerName());
+        } else if (PREMIUM_LIST_MESSAGE.equals(parsedMessage.typeId())) {
+            premiumUsernames.clear();
```
TuxCoding: I'm not sure about this one: I think Velocity uses a event-loop bound to the player connection. That means there could be multiple `PluginMessageEvent` fired concurrently. This would mean newer events might contain old data.

**Message A**: Should update list to: [1, 2 , 3]
**Message B**: Should update list to: [1, 3]

**Program flow:**
**Thread A**: Clear -> premiumUsernames: []
**Thread A**: Add User1 -> premiumUsernames [1]

**Thread B**: Clear
**Thread B**: Add User1 -> premiumUsernames [1]
**Thread A**: Add User2 -> premiumUsernames [1, 2] Old name added
**Thread A**: Add User3 -> premiumUsernames [1, 2, 3]
**Thread A**: Finished

**Thread B**: Add User3 -> premiumUsernames [1, 2, 3] Duplicate ignored
**Thread B**: Finished

Duplicates are ignored, but the list contains the old data from Message A.

authme-core/src/main/java/fr/xephi/authme/service/PremiumService.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    /**
+     * Enables premium mode for the given player. Verifies the player is logged in, that the feature
+     * is enabled, and that the player has an official Minecraft account via the Mojang API.
+     * On success, stores the Mojang UUID as {@code premium_uuid} in the database.
+     *
+     * @param player the player requesting premium mode
+     */
+    public void enablePremium(Player player) {
```
TuxCoding, line 63: FastLogin got complains that players just used the premium command without actually owning a Mojang account. This would effectively lock them out, because we would request online mode authentication from an offline player which the offline client then disconnects from when it sees it. 

AFAIK the current implementation here, does the same thing. Correct me if I'm wrong. Maybe we should add a verification step to prevent such accidents?

authme-core/src/main/java/fr/xephi/authme/listener/packetevents/PremiumVerificationPacketListener.java
```diff
@@ -0,0 +1,212 @@
...
+            handleLoginStart(event);
+        } else if (event.getPacketType() == PacketType.Login.Client.ENCRYPTION_RESPONSE) {
+            handleEncryptionResponse(event);
+        }
+    }
+
+    private void handleLoginStart(PacketReceiveEvent event) {
+        WrapperLoginClientLoginStart wrapper = new WrapperLoginClientLoginStart(event);
+        String username = wrapper.getUsername();
+        if (username == null || username.isEmpty()) {
```
TuxCoding: Adding a check for valid premium usernames would prevent a database lookup here, because this code operates directly on packet level without any safety or bot detection any reduction might be useful.

docs/premium.md
```diff
@@ -0,0 +1,190 @@
...
+---
+
+## Frequently asked questions
+
+**Q: Can I use this on an online-mode server?**  
+A: Online-mode servers already enforce Mojang authentication at the server level — you do not
+need AuthMe's premium bypass at all. AuthMe is primarily designed for offline-mode servers.
+
+**Q: What happens if Mojang's session server is down?**  
+A: `hasJoined` returns an error → the verified UUID is not stored → `canBypassWithPremium()`
```
TuxCoding: If the session servers are down, wouldn't the client also be unable to connect to those servers and terminate the connection?

authme-core/src/main/java/fr/xephi/authme/service/MojangApiService.java
```diff
@@ -0,0 +1,120 @@
...
+    MojangApiService() {
+    }
+
+    /**
+     * Queries the Mojang API to resolve the online UUID for a given username.
+     *
+     * @param username the player name to look up
+     * @return the Mojang online UUID, or empty if the account does not exist or an error occurred
+     */
+    public Optional<UUID> fetchUuidByName(String username) {
```
TuxCoding, line 42: Currently unused, but if we start using it, we should handle `RateLimiting` (429) errors, because this endpoint is very strictly limited.

TuxCoding (changes requested):
I hope you don't mind my many comments :D

authme-core/src/main/resources/messages/messages_ro.yml
```diff
@@ -212,4 +212,23 @@ admin:
...
+  admin:
+    not_registered: '&c%name nu este înregistrat.'
+    already_enabled: '&eModul premium este deja activat pentru %name.'
+    account_not_found: '&cNu a fost găsit niciun cont Mojang pentru %name.'
+    enable_success: '&2Modul premium activat pentru %name.'
+    not_enabled: '&eModul premium nu este activat pentru %name.'
+    disable_success: '&2Modul premium dezactivat pentru %name.'
+    impostor_kicked: '&eUn jucător conectat ca %name a avut un UUID diferit și a fost dat afară.'
+    kick_reason: '&cSetările premium ale contului tău au fost modificate de un administrator. Te rugăm să te reconectezi.'
     first_not_defined: '&cPrimul spawn a eșuat, te rugăm să încearcă să definești primul spawn.'
```
ljacqu: This should come after `not_defined` L214

Xephi: Nice catch

authme-velocity/src/main/java/fr/xephi/authme/velocity/VelocityProxyBridge.java
Xephi: Final verification happen in the backend

authme-velocity/src/main/java/fr/xephi/authme/velocity/VelocityProxyBridge.java
Xephi: Nice catch, made it atomic :D

authme-core/src/main/java/fr/xephi/authme/service/PremiumService.java
Xephi: The UUID v3 path calls fetchUuidByName() and refuses enrollment if no Mojang account exists for that name.
The safest guard would be to only allow /premium enrollment when the player's connection already has a v4 UUID.
Meaning the server or proxy has already verified they own the Mojang account. On a pure offline-mode server with a cracked client, no v4 UUID is possible, so enrollment would be blocked entirely until they prove ownership first.

TuxCoding: I thought about requiring players to rejoin to confirm the change. If the onlinemode authentication is successful, commit/persist the premium state otherwise restore the original state. What do you think?

Xephi: Good idea, i'll check if it's easily do-able

authme-core/src/main/java/fr/xephi/authme/service/bungeecord/BungeeReceiver.java
```diff
@@ -94,6 +97,11 @@ public void onPluginMessageReceived(String channel, Player player, byte[] data)
 
         if (type.get() == MessageType.PROXY_STARTED) {
             logger.info("Proxy plugin '" + argument + "' has started and registered the authme:main channel");
+            java.util.List<String> premiumNames = dataSource.getPremiumUsernames();
```
TuxCoding: Behind the scenes this performs a blocking SQL call on the main thread.

<details>

**Move premium list lookup off the hot path or implement proper caching.**

`dataSource.getPremiumUsernames()` performs a storage-backed SQL query (via `columnsHandler.retrieve()`) synchronously within the `PROXY_STARTED` message handler. If the database is slow or unavailable, this blocks the proxy handshake.

The `CacheDataSource` wrapper exists in the codebase but does not cache this method—it delegates directly. Consider either:
- Caching the premium names list with periodic updates
- Loading it asynchronously on server start and updating it via plugin messaging
- Adding explicit error handling and a fallback at the call site

</details>

authme-core/src/main/java/fr/xephi/authme/service/bungeecord/BungeeSender.java
```diff
@@ -96,8 +98,60 @@ public void sendAuthMeBungeecordMessage(Player player, MessageType type) {
...
+        out.writeUTF(String.join(",", premiumUsernames));
+        bukkitService.sendAuthMePluginMessage(carrier, out.toByteArray());
+    }
+
+    private void sendPremiumNotification(MessageType type, String username) {
+        if (!isEnabled || !plugin.isEnabled()) {
+            return;
+        }
+        Optional<? extends Player> carrier = bukkitService.getOnlinePlayers().stream().findFirst();
+        carrier.ifPresent(p -> {
```
TuxCoding: This fails silently if no player is online. Maybe we should at least log it. For example if the command is executed from the console.

<details>
**Don't drop premium updates when no carrier is online.**

`sendPremiumNotification()` silently returns when the server is empty, so `/premium` and `/freemium` changes can be persisted locally but never reach the proxy. That leaves proxy state stale until some later full resync.

Queue the message until a carrier exists, or at least log/retry so the update is not lost.
</details>

authme-core/src/main/java/fr/xephi/authme/service/PremiumLoginVerifier.java
```diff
@@ -0,0 +1,206 @@
...
+                byte[] decryptedToken = rsaDecrypt(encVerifyToken);
+
+                if (!Arrays.equals(decryptedToken, pend.verifyToken())) {
+                    logger.warning("Verify token mismatch during premium verification for '"
+                        + pend.username() + "'");
+                    return Optional.empty();
+                }
+
+                String serverHash = computeServerHash(sharedSecret);
+                return mojangApiService.hasJoined(pend.username(), serverHash);
```
TuxCoding: `supplyAsync` without a threadpool runs on the JVM common pool which is limited in size. Maybe we should use the bukkit executor for that?

<details>
Lines 130–142 use CompletableFuture.supplyAsync(...) without an executor. The call to mojangApiService.hasJoined(...) is synchronous and performs blocking HTTP I/O with a 5-second timeout. Under login spikes, this saturates the JVM common pool and degrades performance.

Provide a dedicated executor to handle premium verification tasks
</details>

docs/premium.md
```diff
@@ -0,0 +1,192 @@
...
+settings:
+  enablePremium: true
+```
+
+### 3. Enroll players
+
+Players must opt in individually after the admin enables premium mode.
+
+**Player commands:**
+```
```
TuxCoding: Add language barriers to silent markdown warnings


```suggestion
```text
```

TuxCoding (changes requested):
I performed a CodeRabbit review in https://github.com/TuxCoding/AuthMeReloaded/pull/14 and it found some noteworthy things. I copied the relevant things here.

authme-core/src/main/resources/messages/messages_de.yml
```diff
@@ -7,18 +7,18 @@
 registration:
     disabled: '&cRegistrierungen sind deaktiviert'
     name_taken: '&cDieser Benutzername ist schon vergeben'
-    register_request: '&3Bitte registriere dich mit "/register <passwort> <passwortBestätigen>"'
-    command_usage: '&cBenutze: /register <passwort> <passwortBestätigen>'
-    reg_only: '&4Nur für registrierte Spieler! Bitte besuche http://example.com um dich zu registrieren.'
+    register_request: '&3Bitte registriere dich mit "/register <passwort> <passwortBestÃ¤tigen>"'
+    command_usage: '&cBenutze: /register <passwort> <passwortBestÃ¤tigen>'
```
TuxCoding: UTF-8 file encoding is broken now. Looks like other translation files are affected too.

This means umlaute like `öä` no longer work.

authme-core/src/main/java/fr/xephi/authme/service/PendingPremiumCache.java
```diff
@@ -0,0 +1,108 @@
...
+     *
+     * @param name the player name (case-insensitive)
+     * @return the pending Mojang UUID, or null
+     */
+    public UUID getPendingUuid(String name) {
+        PendingEntry entry = pending.get(name.toLowerCase(Locale.ROOT));
+        if (entry == null) {
+            return null;
+        }
+        if (System.currentTimeMillis() > entry.expiresAt()) {
```
TuxCoding, line 67: Race condition if a concurrent thread invokes `addPending` for the name username. It will then remove a fresh copy. Use

`remove(Object, Object)` or lock the struct.

authme-core/src/main/java/fr/xephi/authme/service/bungeecord/BungeeReceiver.java
```diff
@@ -87,13 +100,23 @@ public void onPluginMessageReceived(String channel, Player player, byte[] data)
...
 
         if (type.get() == MessageType.PROXY_STARTED) {
             logger.info("Proxy plugin '" + argument + "' has started and registered the authme:main channel");
+            final String proxyName = argument;
+            final Player carrier = player;
+            bukkitService.runTaskAsynchronously(() -> {
+                List<String> premiumNames = dataSource.getPremiumUsernames();
+                if (!premiumNames.isEmpty()) {
+                    bukkitService.scheduleSyncTaskFromOptionallyAsyncTask(() -> {
+                        bungeeSender.sendPremiumList(carrier, premiumNames);
```
TuxCoding: Consider retrieving a fresh carrier here, because the player could went offline

authme-core/src/main/java/fr/xephi/authme/listener/packetevents/PremiumVerificationPacketListener.java
```diff
@@ -0,0 +1,222 @@
...
+
+        Optional<byte[]> encVerifyTokenOpt = wrapper.getEncryptedVerifyToken();
+        if (!encVerifyTokenOpt.isPresent()) {
+            // Client responded with a signed nonce (1.19.1+ chat signing). We sent a plain verify
+            // token so this should not happen; fall through to password auth without verifying.
+            logger.warning("Player '" + username + "' returned a signed nonce during premium "
+                + "verification (expected plain verify token). Resuming without premium bypass.");
+            loginVerifier.cleanupPending(connectionKey);
+            event.setCancelled(true);
+            resumeLogin(user, username, clientVersion, playerUUID);
```
TuxCoding, line 133: Only a note - not necessary change requested:

The client like expects encryption at this point if the decryption only fails at the server-side. So we might need to kick the player instead - the client wouldn't understand it anyway.

authme-core/src/main/java/fr/xephi/authme/service/bungeecord/BungeeSender.java
```diff
@@ -96,8 +98,76 @@ public void sendAuthMeBungeecordMessage(Player player, MessageType type) {
...
+     * @param carrier         the player used as message carrier
+     * @param premiumUsernames the list of premium usernames (lowercase)
+     */
+    public void sendPremiumList(Player carrier, List<String> premiumUsernames) {
+        if (!isEnabled || !plugin.isEnabled()) {
+            return;
+        }
+        ByteArrayDataOutput out = ByteStreams.newDataOutput();
+        out.writeUTF(MessageType.PREMIUM_LIST.getId());
+        out.writeUTF(String.join(",", premiumUsernames));
```
TuxCoding: This could run into issues if the list is too large like 2k names. Impossible for small servers, but what about larger servers. Considering it is contains all existing premium names from the database. What do you think how we should handle it? It's just a only concern - maybe it's not relevant?

authme-core/src/main/java/fr/xephi/authme/listener/packetevents/PremiumVerificationPacketListener.java
```diff
@@ -0,0 +1,222 @@
...
+        ClientVersion clientVersion = user.getClientVersion();
+        String connectionKey = connectionKey(user);
+        // MC >= 1.20.2 includes the player UUID in LOGIN_START; must be forwarded when re-injecting
+        UUID playerUUID = wrapper.getPlayerUUID().orElse(null);
+
+        // Cancel immediately so the vanilla server does not process this LOGIN_START yet
+        event.setCancelled(true);
+
+        // DB lookup must happen off the Netty event loop to avoid blocking
+        CompletableFuture.runAsync(() -> {
```
TuxCoding: Bukkit run async here too, because this a blocking operation as well.

authme-core/src/main/java/fr/xephi/authme/service/PendingPremiumCache.java
TuxCoding: Although **EXTREMLY** unlikely to happen

TuxCoding (changes requested):
How did the translation files broke? :D

The pending verification looks good so far.

lokspel:
if you are doing /premium then rejoin and login it says you're already logged in
image

lokspel:
and what will happens if the license name will be changed, account will be locked?

Xephi:
> if you are doing /premium then rejoin and login it says you're already logged in image

Should be fixed with ec5038ac057c389c5ac6e17cea925bf451e3f7e9



> and what will happens if the license name will be changed, account will be locked?

During the /premium process, you're kicked, you have to re-login to enable the premium feature completely. If you change your account name later, you'll may have to /login & /freemium + /premium again, but you won't be locked

lokspel:
everything works
    kick_reason: '&cНастройки Premium вашего аккаунта были изменены администратором. Пожалуйста, переподключитесь.'
but the kick reason for premium is not translated if the client is in russia

lokspel:
we still use sessions even tho we dont need if we already use premium mode

authme-core/src/main/java/fr/xephi/authme/service/bungeecord/BungeeSender.java
Xephi: Maybe we could introduce a new premium.list.chunk message type. sendPremiumList will split the list into batches of ~1000 names and send them as sequential chunk messages, each carrying a sequence number and a "last chunk" flag. The proxy accumulates chunks and only replaces its premium set once the final chunk arrives, so there's no window where the cache is empty.
I think 1000 is a good limit, so we stay under 20kb even in the worst case scenario (every name are 16chars).
What do you think ?

Xephi:
> everything works kick_reason: '&cНастройки Premium вашего аккаунта были изменены администратором. Пожалуйста, переподключитесь.' but the kick reason for premium is not translated if the client is in russia

1b9f6476299b0b51e7414a3347b6cd933d7a6528

> we still use sessions even tho we dont need if we already use premium mode

188c752769816d69f5f0caee69e5ea3753d33923

~~wah, can't write a good commit message anymore, time to sleep~~

authme-core/src/main/java/fr/xephi/authme/service/bungeecord/BungeeSender.java
Xephi: @TuxCoding 03632e6ded4897a1f637ed3ee843ed7eef3d6889

lokspel:
Why we store only online uuid in the db? 
I think we should store offline as well in case if someone migrated from another plugin like https://github.com/whereareiam/Identica or nlogin with custom generated uuids

lokspel:
Everything works now feel free to merge

PR #3077, opened 2026-05-24
fix: restore walk/fly speed correctly on quit, repair corrupted speed…

lokspel:
… on join


lokspel:
Fixes: 

https://github.com/user-attachments/assets/540a9da6-9c60-4ad6-883c-b87dc63656b6

Issue #3076, opened 2026-05-23
Incorrect behavior of perform.login when switching between velocity servers


seansolovev:
### What behaviour is observed?

When a player switches from an auth server to another backend server that also has AuthMe installed, AuthMe on the destination server asks for the password again instead of recognizing the player as already authenticated.

AuthMe Velocity does NOT send `perform.login` to `lobby` when the player switches there. The Velocity log shows no `authmevelocity` output at all during the switch to `lobby`. AuthMe on `lobby` receives no login signal and kicks the player after the auth timeout.

### Expected behaviour

Player switches from `survival` to `lobby` without being asked for a password again. AuthMe Velocity should send `perform.login` to `lobby` so the player is automatically authenticated.

### To Reproduce

- **Velocity proxy** with AuthMe Velocity plugin
- **survival** — auth server (listed in `authServers`), AuthMe installed
- **lobby** — non-auth server (not in `authServers`), AuthMe installed

`plugins/authmevelocity/config.yml`:
```yaml
authServers:
  - survival
allServersAreAuthServers: false
autoLogin: true
proxySharedSecret: "<same secret on all servers>"
```

AuthMe `config.yml` on every backend server:
```yaml
Hooks:
  bungeecord: true
  proxySharedSecret: "<same secret as proxy>"
```

`/config/paper-global.yml`:
```yaml
proxies:
  bungee-cord:
    online-mode: true
  proxy-protocol: false
  velocity:
    enabled: true
    online-mode: false
    secret: "<secret>"
```

### Plugin list

`velocity`
```
[05:48:25 INFO]: Plugins: velocity, authmevelocity, luckperms, minimotd-velocity, skinsrestorer, viaversion, sonar, viabackwards
```

`survival`
```
[05:45:04 INFO]: ℹ Server Plugins (21):
[05:45:04 INFO]: Paper Plugins (2):
[05:45:04 INFO]:  - BlueSlimeCore, ToolStats
[05:45:04 INFO]: Bukkit Plugins (19):
[05:45:04 INFO]:  - AuthMe, Chunky, DeluxeMenus, Essentials, EssentialsChat, EssentialsSpawn, HoloMobHealth, HoloMobHealth-PacketEvents, InteractionVisualizer, LuckPerms
[05:45:04 INFO]:  packetevents, PlaceholderAPI, SkinsRestorer, Vault, ViaBackwards, ViaVersion, VillagerInABukkit, WorldEdit, WorldGuard
```

[...]

PR #3085, opened 2026-05-29
Fix Velocity perform.login not reaching target server on cross-server switch

lokspel:
**Problem:** When switching Velocity servers (e.g. survival → lobby), players get prompted to log in again. perform.login never reaches the new server because event.getPlayer().getCurrentServer() returns an empty Optional or the previous server (survival) during ServerConnectedEvent.

**Fix:**
1. Send perform.login via event.getServer() (the RegisteredServer the player just joined) instead of event.getPlayer().getCurrentServer() — always correct, no race condition.
2. Fix infinite retry loop in scheduleRetry(): increment attempt counter before checking getCurrentServer(), so it progresses even when the server connection isn't available yet.

### DIFF
--- a/authme-bungee/pom.xml
+++ b/authme-bungee/pom.xml
@@ -32,6 +32,11 @@
             <scope>compile</scope>
         </dependency>
         <dependency>
+            <groupId>com.github.retrooper</groupId>
+            <artifactId>packetevents-bungeecord</artifactId>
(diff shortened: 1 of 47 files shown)


lokspel:
Fixes probably https://github.com/AuthMe/AuthMeReloaded/issues/3076

PR #3096, opened 2026-06-21
feat(proxy): Added Login/Logout event tracking in the bungeecord/velocity modules

GoodTimes14:
this changes allow other plugin developers to listen for AuthMe logins and logouts on the proxy side, really useful if you want to block certain mechanics that should watch for the current player authentication state.


sgdc3:
Looks good to me 👍🏻

authme-bungee/src/main/java/fr/xephi/authme/bungee/BungeeProxyBridge.java
```diff
@@ -244,11 +246,21 @@ public void onPluginMessage(PluginMessageEvent event) {
...
         if (LOGIN_MESSAGE.equals(parsedMessage.typeId())) {
             if (configuration.isAuthServer(server.getInfo())) {
                 logger.info("Player " + parsedMessage.playerName() + " authenticated on auth server '"
                     + server.getInfo().getName() + "'");
+
+                ProxiedPlayer player = proxyServer.getPlayer(playerName);
+                if (player != null) {
+                    boolean premium = requiresPremiumVerification(parsedMessage.playerName());
+                    proxyServer.getPluginManager().callEvent(new AuthMeBungeeLoginEvent(player, premium));
+                }
```
Xephi, line 260: isPremium() is always false for premium players (non-normalized name passed to premium lookup).
The new code computes boolean premium = requiresPremiumVerification(parsedMessage.playerName()) using the raw player name (e.g. "Alice"), but requiresPremiumVerification checks premiumUsernames / pendingPremiumUsernames, which store lowercase-normalized names (populated via normalizeName(...) at lines 287/306). So for any mixed/upper-case username, the lookup misses and AuthMeBungeeLoginEvent.isPremium() reports false even for genuinely premium players. The normalized playerName variable was just computed three lines above and should be used here (line 249) exactly as the Velocity side does (requiresPremiumVerification(normalizeName(...))). 
Btw, the new test uses a non-premium player, so it doesn't catch this.

Xephi (changes requested):

authme-bungee/src/main/java/fr/xephi/authme/bungee/BungeeProxyBridge.java
GoodTimes14: Fixed problem and added test coverage for premium players too

PR #3101, opened 2026-06-29
Fix player freeze after /freemium with proxy auto-login

lokspel:
### Bug
Player does /freemium, rejoins — can't move. Rejoins again — works fine,
dialog or session message appears.

### Why
After /freemium the backend tells the proxy "this player is no longer premium".
If the player rejoins before the proxy gets that message, the proxy still
sends an auto-login request with the old premium UUID. The backend rejects
it (correctly), but then the code just returned — no login, no limbo,
no dialog, player frozen.

### Fix
When the proxy auto-login request is rejected, don't return early —
continue to session check and limbo/dialog like nothing happened

### Showcase of Bug

Rejoining after the freemium.
image


lokspel:
( i don't know how to fix the prejoin dialog, it doesnt show up too )

lokspel:
> ( i don't know how to fix the prejoin dialog, it doesnt show up too )

resolved.

PR #3119, opened 2026-08-06
feat(api): add forceLoginFromProxy(String) for dialog-safe proxy pre-…

### LINKED ISSUE #3118: Dialog feature API hook
### What feature do you want to see added?

I made a plugin to auto-login players upon certain conditions. This feature works good-ish (it requires some multiple attempts to force login the player) to achieve full login.

This feature stopped working with the dialog update.

Can you please add an API to allow us to bypass the login fully for certain players in the dialog mode?

The hook goes from Velocity -> Bukkit via plugin message and then on player join it does the AuthMe force login method.

On dialog mode, this no longer works.

### Are there any alternatives?

Not that I am aware of.

### Anything else?

_No response_

linsaftw:
…join login

Bypasses Paper/Folia pre-join dialogs when called before Player creation. Queue-first ordering ensures race-safe convergence with AsynchronousJoin. Caller must authenticate the proxy signal; completion is async.

Fixes #3118

### DIFF
--- a/authme-core/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
+++ b/authme-core/src/main/java/fr/xephi/authme/api/v3/AuthMeApi.java
@@ -266,6 +266,27 @@ public class AuthMeApi {
     }
 
     /**
+     * Force a player to login by name, bypassing password verification.
+     * This is intended for trusted proxy integrations where the proxy has already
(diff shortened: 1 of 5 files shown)


linsaftw:
Tested and fully working!

Faboit1:
Bump +1

PR #3128, opened 2026-08-30
Restore 1.8 legacy support

linsaftw:
In collaboration of multiple server owners, we reintroduce 1.8 legacy support.

Still early and might have problems.


linsaftw:
https://drive.google.com/file/d/1mDNGrw6_Eqx1T_LF6QQFhHU5LWFiCdx8/view?usp=sharing

Xephi:
Hey, thanks for your contribution.
We may prefer an another module for these legacy build, so we can keep Fat Jar and Java 8 as far as possible of recents configurations.

lokspel:
I really need 1.8

linsaftw:
Done :)

sgdc3:
@linsaftw I think what Xephi meant was adding a pre-1.16 maven module, not making the 1.16+ version compatible with 1.8

linsaftw:
> @linsaftw I think what Xephi meant was adding a pre-1.16 maven module, not making the 1.16+ version compatible with 1.8



> @linsaftw I think what Xephi meant was adding a pre-1.16 maven module, not making the 1.16+ version compatible with 1.8

Thats what I did.

sgdc3:
> > @linsaftw I think what Xephi meant was adding a pre-1.16 maven module, not making the 1.16+ version compatible with 1.8
> 
> > @linsaftw I think what Xephi meant was adding a pre-1.16 maven module, not making the 1.16+ version compatible with 1.8
> 
> Thats what I did.

Yep sorry, I was distracted when I commented, the actual issue is that much of the "fallback" logic now is inside the core authme module, we would like to keep all the legacy listeners and fallbacks inside the 1.8 module.

lokspel:
i don't like the cores idea tbh.

lokspel:
i like when it was universal jar

linsaftw:
> > > @linsaftw I think what Xephi meant was adding a pre-1.16 maven module, not making the 1.16+ version compatible with 1.8
> > 
> > 
> > > @linsaftw I think what Xephi meant was adding a pre-1.16 maven module, not making the 1.16+ version compatible with 1.8
> > 
> > 
> > Thats what I did.
> 
> Yep sorry, I was distracted when I commented, the actual issue is that much of the "fallback" logic now is inside the core authme module, we would like to keep all the legacy listeners and fallbacks inside the 1.8 module.

What do you mean? The extra protective layer listeners? Those are not legacy and unrelated to the 1.8 module. 

They are being extracted to be able to register the PlayerListener without having to entirelly recode something that is already good. Because the PlayerListener is a massive meta-class that contains all listeners. This poor design leds to do this. What do you want t o do instead? To make another meta-class listener but for legacy? I think its redundant.

Issue #3131, opened 2026-09-03
Zero day login exploit


nomadanarchy:
### What behaviour is observed?

https://www.youtube.com/watch?v=qeNKXSnaOBg

### Expected behaviour

https://www.youtube.com/watch?v=qeNKXSnaOBg

### To Reproduce

https://www.youtube.com/watch?v=qeNKXSnaOBg

### Plugin list

https://www.youtube.com/watch?v=qeNKXSnaOBg

### Server Implementation

BungeeCord

### Database Implementation

SQLite

### AuthMe Version

https://www.youtube.com/watch?v=qeNKXSnaOBg

### Configuration

https://www.youtube.com/watch?v=qeNKXSnaOBg

TuxCoding:
Source identified. Working on a fix.

NikOverflow:
@TuxCoding I wrote stuff here: https://github.com/AuthMe/AuthMeReloaded/issues/3132 It is not only an authme issue.

Xephi:
Fixed in #3135

PR #3133, opened 2026-09-03
[Security] Bind dialog requests to the InetSocketAddress

TuxCoding:
# Background

Multiple clients with the same username/uuid are allowed in the configuration phase (part of the dialog system). Only in later stages players are getting kicked with "logged in from another location".

This means the UUID and names we use as [identifiers](https://github.com/AuthMe/AuthMeReloaded/blob/f5afba06d8b973d415034e22bfff38911b3716b0/authme-paper-common/src/main/java/fr/xephi/authme/listener/PaperDialogFlowListener.java#L55) there are not unique. Later connection requests will [remove/override](https://github.com/AuthMe/AuthMeReloaded/blob/f5afba06d8b973d415034e22bfff38911b3716b0/authme-paper-common/src/main/java/fr/xephi/authme/listener/PaperDialogFlowListener.java#L108) earlier associations to the ongoing login attempt. This has the consequence that if an earlier legimate login request enters the password correctly, only the latest open dialog will be approved, which could be an attacker.

# Related

#3131

# Reproduce

0. Default config
1. Join two clients
2. Enter password in the first client
3. Second (incorrect) client is approved
4. First client (the legimate one) gets a timeout

## Solution idea

**Draft only**

Bind the open login dialog to the connection (player object or connection object) which makes it unique per connection. This is what this PR does. **In its current state it fixes the issue for single paper server setup. Force logins from proxies or commands are disabled, because it needs a way to differentiate multiple connections instead of relying only the player name.**

### DIFF
--- a/authme-core/src/main/java/fr/xephi/authme/process/join/AsynchronousJoin.java
+++ b/authme-core/src/main/java/fr/xephi/authme/process/join/AsynchronousJoin.java
@@ -153,11 +153,11 @@ public class AsynchronousJoin implements AsynchronousProcess {
         String name = player.getName().toLowerCase(Locale.ROOT);
         String ip = PlayerUtils.getPlayerIp(player);
         UUID playerId = player.getUniqueId();
-        String pendingLoginPassword = preJoinDialogService.consumePendingLoginPassword(playerId);
-        String pendingRecoveryEmail = preJoinDialogService.consumePendingRecoveryEmail(playerId);
+        String pendingLoginPassword = preJoinDialogService.consumePendingLoginPassword(player.getAddress(), playerId);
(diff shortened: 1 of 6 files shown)


lokspel:
https://github.com/IndrajeethY/AuthMeReloaded/tree/fix/prejoin-session-hijack

TuxCoding:
~~@lokspel That's an incorrect hack, because it only creates an additional password dialog if the password was incorrect.~~

Sorry looked only at the latest commit. 

EDIT: Actually it is the same idea.

Xephi:
https://github.com/AuthMe/AuthMeReloaded/pull/3135

PR #3134, opened 2026-09-03
fix(proxy): route authentication to the originating player

moreveal:
Login completion was routed by player name after the backend had already sent the message through the player who entered the password. During overlapping same-name connections, the proxy lookup could resolve the older player and move that session instead.

Keep the player attached to the plugin message through the authentication flow. Bungee uses the message receiver and Velocity uses the message target. Both reject session messages whose payload name does not match that player.

Authentication state, pending login retries, ACK handling, redirects, and disconnect cleanup now belong to the exact connection. The backend cache also tracks the Bukkit Player that owns the authenticated session, so an old disconnect cannot clear its replacement.

Added regression coverage for concurrent same-name connections, mismatched message targets, stale disconnects, backend message carriers, and pending login routing.

### DIFF
--- a/authme-bungee/src/main/java/fr/xephi/authme/bungee/BungeeAuthenticationStore.java
+++ b/authme-bungee/src/main/java/fr/xephi/authme/bungee/BungeeAuthenticationStore.java
@@ -2,35 +2,26 @@ package fr.xephi.authme.bungee;
 
 import net.md_5.bungee.api.connection.ProxiedPlayer;
 
-import java.util.Locale;
-import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.IdentityHashMap;
+import java.util.Map;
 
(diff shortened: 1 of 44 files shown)


lokspel:
What PR fixes exactly? Can you link the issue?

moreveal:
> What PR fixes exactly? Can you link the issue?

No GitHub issue - this is an independent PR based on the vulnerability demonstrated in this video:
[YouTube demonstration](https://www.youtube.com/watch?v=qeNKXSnaOBg)

lokspel:
has been fixed in https://github.com/AuthMe/AuthMeReloaded/pull/3135

moreveal:
#3135 fixes the reported exploit earlier in the connection lifecycle by reserving the player name per connection and keeping pre-join state per session.

This PR hardens later proxy routing, but I don't have a separate reproduction against current master. Closing as superseded.

PR #3135, opened 2026-09-03
Claim pre-join connection and re-key it by a session identifier (instead of player's uuid)

### LINKED ISSUE #3131: Zero day login exploit

### LINKED ISSUE #3132: Major AuthMe ForceOp Exploit (nLogin wasn't affected)
0 Day AuthMe login bypass exploit found by @unknown-user
also i think this is the 1st major exploit found with minecrafts new like custom server menus 

basically u join as the other person does and when they enter their pw it lets u in instead of them 😭
you can macro this and just like wait for them just leave join macro
https://www.youtube.com/watch?v=qeNKXSnaOBg

Image

Xephi:
May fix #3132 fix #3131 
Need a strong test before merging

@TuxCoding

### DIFF
--- a/authme-paper-common/src/main/java/fr/xephi/authme/listener/PaperLoginValidationListener.java
+++ b/authme-paper-common/src/main/java/fr/xephi/authme/listener/PaperLoginValidationListener.java
@@ -28,8 +28,8 @@ public class PaperLoginValidationListener implements Listener {
 
     private static final LegacyComponentSerializer LEGACY_SERIALIZER = LegacyComponentSerializer.legacySection();
 
-    /** Extra time a name stays claimed on top of the configured login / register timeout. */
-    private static final long CLAIM_GRACE_MILLIS = TimeUnit.SECONDS.toMillis(10);
+    /** Safety net for a connection that never reports itself as gone; not the release mechanism. */
+    private static final long CLAIM_GRACE_MILLIS = TimeUnit.SECONDS.toMillis(60);
 
     @Inject
     private OnJoinVerifier onJoinVerifier;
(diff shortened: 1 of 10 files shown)


authme-paper-common/src/main/java/fr/xephi/authme/listener/PaperLoginValidationListener.java
```diff
@@ -63,6 +73,70 @@ public void onPlayerServerFullCheck(PlayerServerFullCheckEvent event) {
...
+        }
+    }
+
+    private void verifyConfigurationPhase(PlayerConnectionValidateLoginEvent event,
+                                          PlayerConfigurationConnection connection) {
+        PlayerProfile profile = connection.getProfile();
+        String playerName = profile == null ? null : profile.getName();
+        // No claim means an online player being reconfigured, which must not be checked against itself
+        if (playerName == null || !pendingConnectionRegistry.holdsClaim(playerName, connection)) {
+            return;
```
TuxCoding, line 125: So this means the claim stale or held by another connection? If I understand this correctly, then shouldn't we kick the player then? Otherwise the client could still finish configuring the connection and join the server.

TuxCoding (approved):
I tested the changes. It will fix the mentioned issue. 

In comparison to the other solutions this will limit the dialog flow to a single connection. The downside is that legitimate players could be locked out until the connection is available. It's more restrictive, but offers a cleaner implementation - which is more important IMHO.

authme-paper-common/src/main/java/fr/xephi/authme/listener/PaperLoginValidationListener.java
Xephi: Kicking on every !holdsClaim would break the workflow, connection with no claim never went through the login phase, so it's an online player being reconfigured, and checkSingleSession would kick him over his own presence.

Fixed in 0ef6fc4 , deny the connection when the name is held by another live connection, keep the early return only when there is no claim at all.

Thanks for your review on this !

PR #3139, opened 2026-09-05
Fix single session check to match names case-insensitively

HEBEI77:
checkSingleSession used getPlayerExact which is case sensitive, but auth is looked up by lowercase name and offline UUIDs differ by case.

That lets Steve and steve stay online at the same time on the same account. When one logs in the other gets unlocked too.

Match the javadoc and scan online players with equalsIgnoreCase.


lokspel:
Server isn't actually allowing the same name in the server tho? Tho patch might be useful

HEBEI77:
Good point for the exact same string, the server kicks that itself.

This is for casing. Offline UUIDs are case sensitive so Steve and steve both stay online but map to one AuthMe account. Thats the case the javadoc already claims to cover.

Issue #3122, opened 2026-08-10
"registerTimeout: 0" — breaks registration


Roper89:
### What behaviour is observed?

<img width="678" height="117" alt="Image" src="https://github.com/user-attachments/assets/004b54ed-eccd-443a-900f-e470d9b03298" />

Using the parameters "loginTimeout: 0" and "registerTimeout: 0" prevents logging into the server. Instead of disabling the timer, this works as if it were waiting for 1 second.

### Expected behaviour

In the old version of AuthMeReloaded v5... this worked like disabling the timer during which you had to log in or register.

### To Reproduce

1. Set the "loginTimeout" and "registerTimeout" parameters to 0
2.  Register as a new member.

### Plugin list

AuthMe, CMI, CMILib, FastAsyncWorldEdit, InfernalMobs, LuckPerms, mcMMO, ProtocolLib, WorldEdit, WorldGuard

### Server Implementation

Standalone server (no proxy)

### Database Implementation

SQLite

### AuthMe Version

AuthMeReloaded v6.0.1-SNAPSHOT (build: 2763)

### Configuration

https://gist.github.com/Roper89/874031ffbd2586cb96106ac287c1297a

lokspel:
Confirmed.

lokspel:
it is dialog only issue.

PR #3141, opened 2026-09-05
fix(dialog): respect loginTimeout/registerTimeout=0 as disabled

lokspel:


lokspel:
Tried to solve https://github.com/AuthMe/AuthMeReloaded/issues/3122

